PR #13, opened 2017-10-19
Add default argument for pandas.df

wiso:
In this way it is possible to dump the whole TTree into a pandas
dataframe. The name of the branches are taken from self.branchnames.
No type information is used.


jpivarski:
Sorry for not seeing this earlier. All this time, I'd been assuming that I was automatically watching my own repositories. Now I find that I have to turn this on explicitly.

PR #59, opened 2018-03-07
Use hashlib.md5 to do hash for diskcache

asymmetry:
Since hash() is salted randomly after python3.3, the _index() function in diskcache.py line 637 may not work properly between processes. Although people could manually set PYTHONHASHSEED env to avoid this problem, it might still be good to move to module hashlib.


jpivarski:
Wow, thanks! This was an important thing to know.

PR #62, opened 2018-03-08
Temp fixed a problem preventing JaggedArray from being pickled

asymmetry:
This is probably another tiny bug about the diskcache. It seems the JaggedArray class is not able to be pickled if generated by reading a rootfile (ok if from lists). I tested it with HZZ.root and some codes like this:
```
_disk_cache = uproot.cache.DiskCache.create(50 * 1024**3, _cache_dir)
_cache = uproot.cache.MemoryCache(8 * 1024**3, spillover=_disk_cache, spill_immediately=True)
tree = uproot.open('HZZ.root')['events']
muon_px = tree.arrays(["Muon_Px"], cache=_cache)
```
and the output is:
```
_pickle.PicklingError: Can't pickle <class 'uproot.rootio.TLeafI'>: attribute lookup TLeafI on uproot.rootio failed
```

Digging into the code, it seems that the leafcount attribute in JaggedArray is causing this problem since it saves a dynamically generated class. And pickle does not work very well with dynamic things. The only place leafcount is used is the JaggedArray.aligned() function and after a grep, I did not see any other place this function is called. So I removed this leafcount attribute as a temp work-around and aligned() can still be calculated by comparing starts and stops.

I might be wrong if this function is added on purpose. If so, just drop off this PR and fix it with a more decent way.

uproot is a great project! Thanks a lot!

### DIFF
--- a/uproot/interp/jagged.py
+++ b/uproot/interp/jagged.py
@@ -136,10 +136,7 @@ class asjagged(Interpretation):
         starts = offsets[:-1]
         stops  = offsets[1:]
         content = self.asdtype.finalize(destination.content, branch)
-        leafcount = None
-        if len(branch.fLeaves) == 1:
-            leafcount = branch.fLeaves[0].fLeafCount
-        return JaggedArray(content, starts, stops, leafcount=leafcount)
+        return JaggedArray(content, starts, stops)
 
 def asstlvector(asdtype):
     return asjagged(asdtype, skip_bytes=10)
@@ -215,7 +212,7 @@ class JaggedArray(object):
 
         return JaggedArray(content, starts, stops)
 
-    def __init__(self, content, starts, stops, leafcount=None):
+    def __init__(self, content, starts, stops):
         assert isinstance(content, numpy.ndarray)
         assert isinstance(starts, numpy.ndarray) and issubclass(starts.dtype.type, numpy.integer)
         assert isinstance(stops, numpy.ndarray) and issubclass(stops.dtype.type, numpy.integer)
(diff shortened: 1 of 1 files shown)


jpivarski:
I'm putting in a different fix.

jpivarski:
JaggedArrays with the same `starts` and `stops` are very frequently combined, and we can determine if this is legal (if the two jagged arrays have the same jagged shapes) without walking over all events for all pairs of arrays we might want to combine by verifying that they came from the same counter in ROOT.

Your solution is technically correct, but simply removes the optimization. (I'm touchy about everything that happens for all events and a quadratic number of branches.)

[Release 2.8.6](https://github.com/scikit-hep/uproot/releases/tag/2.8.6) keeps the optimization until the object is pickled, and then we lose track of its origin and have to spend the cycles checking.

jpivarski:
Thanks, though!

asymmetry:
Thanks. Now I understood why this function is designed. I am thinking of __getstate__ and __setstate__ just now but not sure if I could use them correctly. Probably an issue is better for this case. Digging into bugs and try to fix it is fun to me and could drive me to learn more and I did learn something from you solution. Thanks a lot!

jpivarski:
These methods are just overriding the pickle behavior of this object. I made the `leafcount` member transient.

I haven't gone through and surveyed which types are pickleable, so there might be more like this.

anerokhi:
> I haven't gone through and surveyed which types are pickleable, so there might be more like this.

Indeed, `PicklingError("Can't pickle <class 'uproot.rootio.TH1I'>: attribute lookup uproot.rootio.TH1I failed",)`

Issue #68, opened 2018-03-29
Encoding problem when installing


guiguem:
**Issue:**
In Debian Jessie with python 3 installed, installing uproot fails with the following error:
```
>>pip install uproot

Collecting uproot
  Downloading uproot-2.8.12.tar.gz (9.7MB)
    100% |################################| 9.7MB 177kB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-wl1a7pfo/uproot/setup.py", line 74, in <module>
        long_description = get_description(),
      File "/tmp/pip-build-wl1a7pfo/uproot/setup.py", line 44, in get_description
        description = open("README.rst").read()
      File "/build/lib/python3.6/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 595: ordinal not in range(128)
```
This error does not happen on my local machine (Mac in a virtual environment).
I believe the error comes from the fact that the encodings are different.

**Solution:**
The solution I found consists in being explicit in the encoding to use.
Basically, changing https://github.com/scikit-hep/uproot/blob/master/setup.py#L44 
into 
```
description = open("README.rst",'rb').read().decode('utf8', 'ignore')
```
solves the problem in my docker (and also on my Mac).
Should be an easy fix...

jpivarski:
Fixed by #69.

PR #69, opened 2018-03-29
Fixing encoding

guiguem:
Fix Issue #68


jpivarski:
Thanks! I'm surprised this wasn't noticed until now. We've been deploying on Python 3 for a while now.

But anyway, the modified version ought to work, too, and it's more explicit. I like it.

guiguem:
Perfect! thanks for the excellent work btw!

PR #75, opened 2018-04-28
Object not in std::vector

reikdas:


jpivarski:
You need to `git add issue72.root`

reikdas:
Oops I seemed to have forgotten to rename the file.

jpivarski:
`git mv` so that the file size doesn't count against us twice...

reikdas:
I just checked the files uploaded. issue74.root is only present once.

uproot/interp/auto.py
```diff
@@ -192,6 +194,14 @@ def transform(node, tofloat=True):
             if len(branch.fBranches) > 0 and all(len(x.fLeaves) == 1 and x.fLeaves[0].fLeafCount is branch.fLeaves[0] for x in branch.fBranches):
                 return asdtype(">i4")
 
+            if isinstance(branch._streamer, uproot.rootio.TStreamerObject):
+                for x in all_classes:
+                    if (branch._streamer.fTypeName == x._classname):
+                        try:
+                            return asobj(x, branch._context)
+                        except:
```
jpivarski: This `try-catch` is too wide— hiding too many exceptions will make it difficult to debug later. What exception exactly are you trying to catch? Under what conditions does it happen? Can you do `if` statements for that instead, so that surprising exceptions still get raised?

If you're copying something I did, please point out my bad example.

tests/test_issues.py
```diff
@@ -136,3 +136,11 @@ def test_issue66(self):
     def test_issue70(self):
         f = uproot.open("tests/samples/issue70.root")
         self.assertEqual(f.keys(), [])
+
+    def test_issue74(self):
+        f = uproot.open("tests/samples/issue74.root")['Events']
+        self.assertTrue(isinstance(y, uproot.rootio.TVector3) for x in f.array('bees.xyzPosition') for y in x)
```
jpivarski: How is it that `'bees.xyzPosition'` is a list (`x`) of `TVector3` (`y`) and then in the next check, each has a `fX`, `fY`, `fZ`?

jpivarski:
Very nice overall! I just have a few requests.

tests/test_issues.py
reikdas: Could you please elaborate a bit? In the fX, fY, fZ check, I am just iterating over the contents of bees.xyzPosition. And the tests seem to be working. Did you mean I am already specifying x here and then using the same x to iterate in the next check?

jpivarski:
I couldn't find this comment in GitHub anymore, so I'll answer it here.

The test confuses me because I can't understand how both can be right. The
first seems to be verifying that the bees.xyzPosition is a list of TVector3
(per event) and the second seems to be verifying that it is a TVector3
object (per event). Sure, the test is succeeding, so I'm reading it wrong,
but how am I reading it wrong? Does the interface somehow allow the user to
access the same data both as a TVector3 object and also as a (single
element?) iterable of TVector3? If so, please eliminate the later access
method.

(Tests not only have to succeed, they have to be understandable, so that
when they succeed, you know the code they're testing is right. It's
possible to have succeeding tests with the tests proving that the code is
wrong, but perhaps obliquely hiding that fact.)

Jim

tests/test_issues.py
kreczko: If you rewrite the code you have, it becomes more clear were the confusion is coming from.
lines 142-146 are logically identical to
```python
for x in f.array('bees.xyzPosition'):
    for y in x:
        # y is a TVector3
        self.assertTrue(isinstance(y, uproot.rootio.TVector3)
   # further down
    self.assertEqual(x.fX, 1.0) # x is also TVector3?
```
```

PR #77, opened 2018-05-03
Able to read TStrings

reikdas:


reikdas:
Could you indicate the line number? If you are talking about line 262 in auto.py, there is nothing under that. Line 264 is on a different indentation.

jpivarski:
Yes, this works. I downloaded and tested the code, then compared the result against what C++ ROOT says for the same data (`TTree::Scan`). They agree, though the data are unexpected (the numbers don't increase the same way).

We've introduced a distinction in uproot between strings (`std::string` and `char*`) and `TStrings`, whereas I would have liked to treat them all the same way, giving the user nothing but Python strings. But I guess this is alright.

Great work! Thanks a lot!

PR #80, opened 2018-05-14
Fixed basket error

### LINKED ISSUE #79: Problem reading branches with more than one basket on Windows
Hello,
I'm having problems to read any branches that have more than one basket on Windows. The same files are read without any problems on *nix or mac systems.

Here is a minimal example code and the output that illustrates a typical behaviour:
```python
import uproot
file = uproot.open('events.root')
events = file["events"]
pt = events["pt"].array()
```
Output:
```
  File "read_events.py", line 7, in <module>
    pt = events["pt"].array()
  File "C:\Users\konst\AppData\Roaming\Python\Python36\site-packages\uproot\tree.py", line 1202, in array
    self._tryrecover()
  File "C:\Users\konst\AppData\Roaming\Python\Python36\site-packages\uproot\tree.py", line 1490, in _tryrecover
    self._recover()
  File "C:\Users\konst\AppData\Roaming\Python\Python36\site-packages\uproot\tree.py", line 1486, in _recover
    raise ValueError("entries in recovered baskets (offsets {0}) don't add up to total number of entries ({1})".format(entryoffsets, self.numentries))
ValueError: entries in recovered baskets (offsets [0, 288691]) don't add up to total number of entries (1000000)
```
Here is a link to the input events.root used above:
https://www.dropbox.com/s/p480trum1l3f99z/events.root?dl=0

uproot version:  2.8.21

reikdas:
Fixes #79 
Credits go solely to Jim for figuring out the correct solution.


jpivarski:
I just cleaned up the tests a little and bumped the version number. Each change that we want to publish to PyPI must at least increase the revision number (zz in `xx.yy.zz`) but some are big enough to increase the minor number (yy in `xx.yy.zz`). The file-writing feature you'll add this summer will increase the major number from 2 to 3.

PR #81, opened 2018-05-17
Added path to CompressedSource

reikdas:
All credit goes to Jim for the code introduced.


jpivarski:
You don't need to assign credit.

However, maintaining this formality of committing through pull requests, no matter how trivial the change, is a good idea (at least for now).

PR #82, opened 2018-05-18
leaflist dtype interpretation wasn't Python 2 safe; fixed now

jpivarski:


jpivarski:
This is too minor of a change to wait. Of course it will work on Windows; it's a pure Python thing.

PR #83, opened 2018-05-18
Some branches with streamer TStreamerObject cannot be read yet

reikdas:


jpivarski:
No, we can't have a whitelist of acceptable class names. We don't know for certain that there aren't other classes that work— you're excluding a potentially infinite set of non-broken cases.

More importantly, consider what happens whether this pull request is accepted or not:

   * Accepted: users with new classes find that they can't read them in uproot because the Interpretation is `None`. This means, "not implemented _yet_." So they give up on uproot and wait for it to be implemented. They wait forever because we don't know about their use-case and don't even know that they're waiting.
   * Not accepted: by providing an Interpretation of the user's branch type, we're making a statement that we support it. The user tries to use it and it fails immediately with an exception. It looks like a bug and they report it, explaining their use-case and giving us a sample file. Then, not only do we know that something needs to be improved in principle, but we also know that someone cares about it in practice, and they've even giving us an example, which is essential for understanding the issue.

Not giving users error messages is good for public relations— it makes the software feel more stable— but it's bad for progress. Progress is more important.

Keep in mind that I'm not suggesting we give the users buggy software, software that returns the wrong answer (the worst possible case!). This deserialization error results in an immediate exception, which is good. In both bullet points above, the user doesn't get the wrong answer— the software doesn't _wrongly appear_ to do what they want— but in the first bullet point, it encourages the user to give up, while the second encourages the user to help us improve

reikdas:
Should this line be removed - https://github.com/scikit-hep/uproot/blob/master/uproot/interp/auto.py#L133 ?

jpivarski:
Yes, it should be. Somehow I missed that in a code review.

PR #93, opened 2018-06-22
Issue 92

### LINKED ISSUE #92: Two Dimensional Histograms
When I open a 2D histogram in uproot and convert the data to a numpy array, the output only contains bin edges for one axis and under/overflow bins are included in the content, e.g.,:

```
# h2 is the 2d histogram from a root file with 4 x-bins and 2 y-bins
>>> h2.numpy
(array([0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.08, 0.06, 0.03, 0.02, 0.  ,
       0.  , 0.01, 0.04, 0.07, 0.1 , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ]), array([0., 1., 2., 3., 4.]))
```

The bins with 0 are somewhat consistent with the under/overflow bins (in ROOT the 'real' bins start at bin numbers 7-10; 13-16).  However, from the information returned by the `numpy` method, it isn't clear (without already knowing) where the 'real' bins begin and end.

Could the [TH1Methods](https://github.com/scikit-hep/uproot/blob/master/uproot/hist.py) class be extended to support 2D histograms such that the above output would instead be something like:
```
>>> h2.numpy
(
array( [0.08, 0.06, 0.03, 0.02],    # first y-bin
       [0.01, 0.04, 0.07, 0.1] ] ), # second y-bin
array( [ [0., 1., 2., 3., 4.], [0., 1., 2.] ]) . # (x-bins, y-bins) 
)
```

NB: I believe I can recreate what I want using the `allvalues`, `fXaxis`, `fYaxis`, `f*axis.fXmin`, `f*axis.fXmax`, and `f*axis.fNbins` methods. If the bins have variable widths, it appears the `f*axis.fXbins` method will contain the proper bin edges.

demarley:
Closes #92 

Extended the interface for 2d histograms.  Can be tested with the example [file](https://github.com/scikit-hep/uproot/blob/master/tests/samples/hepdata-example.root).

The `show()` method behaves in a similar way as before, but there are line breaks for each y-bin.
The `numpy` method returns the 2d array for the non-under/overflow bins with bin edges for x and y.

Please let me know if you have any feedback -- I tried to match the existing style and such, but if something isn't correct I'm happy to modify it.

### DIFF
--- a/uproot/hist.py
+++ b/uproot/hist.py
@@ -269,6 +269,340 @@ class TH2Methods(TH1Methods):
     def hello(self):
         return "world"
 
+    @property
+    def numbins(self):
+        return self.numXbins * self.numYbins
+
+    @property
+    def numXbins(self):
+        return self.fXaxis.fNbins
+
+    @property
+    def numYbins(self):
+        return self.fYaxis.fNbins
+
+    @property
+    def low(self):
+        return (self.xlow,self.ylow)
+
+    @property
+    def high(self):
+        return (self.xhigh,self.yhigh)
+
+    @property
+    def xlow(self):
+        return self.fXaxis.fXmin
+
+    @property
+    def xhigh(self):
+        return self.fXaxis.fXmax
+
+    @property
+    def ylow(self):
+        return self.fYaxis.fXmin
+
+    @property
+    def yhigh(self):
+        return self.fYaxis.fXmax
+
+    @property
+    def underflows(self):
+        """Underflow values for (x,y) :: xuf[0] == yuf[0] """
+        uf = numpy.array(self.allvalues)
+        xuf = uf[:,0]
+        yuf = uf[0]
+        return (xuf,yuf)
+
+    @property
+    def overflows(self):
+        """Overflow values for (x,y) :: xof[-1] == yof[-1] """
+        of = numpy.array(self.allvalues)
+        xof = of[:,-1]
+        yof = of[-1]
+        return (xof,yof)
+
+    @property
+    def values(self):
+        """Return all values that aren't overflow/underflow"""
+        va = numpy.array(self.allvalues)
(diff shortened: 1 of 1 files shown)


jpivarski:
This looks fantastic, thank you! You did match the surrounding style well. Just a few requests from the API police:

   * No capital letters in method names, like `numXbins` and `numYbins`; it impedes speed typing.
   * Since this is the first two-dimensional object, you're setting a convention about where to put the `x` and `y`. I'd be happy with putting them at the beginnings of words (e.g. `xlow`, `yhigh`), except that when this combines with the `num` convention, you get `numxbins` and `numybins`. When navigating with move-forward-word and move-backward-word keystrokes, that important letter is buried in the middles of the words, making it hard to change quickly. However, religiously putting `x` and `y` at the _ends_ of words leads to gross words like `indexx` (below). How about this? `x` and `y` go at the beginnings of words and this convention trumps `num`: `xnumbins` and `ynumbins`.
   * `underflows` and `overflows` returning tuples: could these be four functions instead: `xunderflow` (singular, even though it's an array), `yunderflow`, `xoverflow`, and `yoverflow`?
   * Same for `xinterval` and `yinterval`. I don't like a switch based on a string because the string can take on infinitely many values, one of which is mapped to "x" and all the others are mapped to "y". It's too much freedom for something that has only two states. And anyway, everything else is functions with names that end in "x" or "y", so it would be more consistent if this were, too. If you want to share code between the implementations, you can introduce an internal method (starting with "_").
   * Same for `xindex` and `yindex`.

jpivarski:
It looks like this is going to merge into my "issue-92" branch, rather than "master," so I'll go ahead and merge it and make the suggested changes myself. Just wanted you to know why I'm making them.

Thanks for this work!

PR #94, opened 2018-06-22
Issue 92

jpivarski:
New 2d histograms.

### DIFF
--- a/uproot/hist.py
+++ b/uproot/hist.py
@@ -172,15 +172,15 @@ class TH1Methods(object):
 
     def __add__(self, other):
         if not isinstance(other, TH1Methods) or self.numbins != other.numbins or self.low != other.low or self.high != other.high:
-            raise TypeError("TH1 histograms can only be combined with other TH1 histograms")
-        return hist(self.numbins, self.low, self.high, name=(self.name if self.name is not None else other.name), title=(self.title if self.title is not None else other.title), values=[x + y for x, y in zip(self.values, other.values)])
+            raise TypeError("TH1 histograms can only be combined with other TH1 histograms with the same binning")
+        return hist(self.numbins, self.low, self.high, name=(self.name if self.name is not None else other.name), title=(self.title if self.title is not None else other.title), allvalues=(numpy.array(self) + numpy.array(other)))
 
     def __radd__(self, other):
         return self.__add__(other)
 
     def __iadd__(self, other):
         if not isinstance(other, TH1Methods) or self.numbins != other.numbins or self.low != other.low or self.high != other.high:
-            raise TypeError("TH1 histograms can only be combined with other TH1 histograms")
+            raise TypeError("TH1 histograms can only be combined with other TH1 histograms with the same binning")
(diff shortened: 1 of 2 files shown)


uproot/hist.py
```diff
@@ -319,3 +678,33 @@ def hist(numbins, low, high, name=None, title=None, values=None, allvalues=None,
         out.fillall(filldata)
 
     return out
+
+def hist2d(xnumbins, xlow, xhigh, ynumbins, ylow, yhigh, name=None, title=None, values=None, allvalues=None, xfilldata=None, yfilldata=None):
+    out = TH1()
```
sbinet: apologies for the drive-by comment, but:

`TH1()` ?
shouldn't this be `TH2()` ?

jpivarski: Nice catch, thanks!

demarley:
Hey @jpivarski thanks for your feedback and merging the previous PR!
I notice that you still left `overflows`, `underflows`, `index`, and `interval`.  Were you planning to remove those methods or change what they return?

I agree that having `index` and `interval` require a string to identify which axis is not ideal.  I can't think of a better way to define them, unfortunately.
One (more complicated) idea that would require some changes would involve adding `index` and `interval` to `TAxis` so that you can do something like:
```
xaxis = self.fXaxis
ind = xaxis.index(data)  # bin index for this data value
inv = xaxis.interval(ind)  # bin edges for this index
```
But this might be a bit beyond what's needed now.

jpivarski:
The reason I went back on that is because TH2 inherits from TH1, so if we don't overload it, the method would still be there and have one-dimensional functionality. THAT would be confusing. I still added the specialized `xinterval` and `yinterval`, which would probably be more often used.

The string-based selector for x and y is still there, but if the string isn't exactly "x" or exactly "y", it's an error. This prevents people from writing "X" (or an equally reasonable but unthought-of string) and getting y.

If there anything you want to a add to `hist2d` to make it usable on it's own? I've only stubbed it out with NotImplementedErrors.

demarley:
I can fill in the `hist2d(...)` real quick with something similar to `hist(...)`.

demarley:
Okay, I added some bits to `hist2d()` and from my (simple & local) tests everything worked like I expected.  Let me know if something is still off and I can fix it.

jpivarski:
Great! If you did some tests constructing `hist2ds` on their own, without pulling them from ROOT files, that's good enough for me. I put the NotImplementedErrors there because I didn't want to implement index manipulations without putting sufficient thought into it.

If you're happy with this, I'll merge and close. Would you like to stay on as a contributor or have me take you off the list (and possibly re-add later) for tidiness?

demarley:
Great! I only put marginally more thought into it as I was mainly interested in the `numpy` feature for what I want to do right now.
I'm happy to stay on as a contributor for at least a little while until `hist2d` is completely vetted by other users.

jpivarski:
Okay, we'll give it a week. Meanwhile, I'm publishing this new release. Thanks again!

PR #98, opened 2018-07-09
Should fix #96

### LINKED ISSUE #96: Support for TLorentzVector not in std::vector?
I am trying to read TLorentzVector type branches using uproot, and in tree.show() I see None as the type (and when doing tree['branchName'].array() the error is: "cannot interpret branch 'branchName' as a Python type." However, I also have a std::vector of TLorentzVectors which can be read.

After reading issue #57 I am thinking there is just not support for reading TLorentzVectors that are not in vectors. If there is a way around this issue it would be greatly appreciated!

reikdas:


reikdas:
The ROOT file is about ~5 MB I think.
It worked previously, because we were just looking at TStreamerObject to have the asobj instances(and TStreamerObjectPointer).

jpivarski:
There might be other cases, like `TStreamerObjectAny` or something like that. It would be unwise to open that up without an example. (Who knows? Maybe `TStreamerObjectAny`'s serialization has an extra byte at the beginning? It's better to not claim we can read it than to claim we can and then do so incorrectly.)

jpivarski:
One AppVeyor done: good enough. The second one is a formality because this branch only differed from master by the one change, so the push request test is literally the same as the branch test. (Because no one else changed master while you were working on this branch.) So I'm going to merge, increase the version number, and deploy to PyPI.

imruffin:
Hey, the new version is working for me, thanks so much!

Issue #100, opened 2018-07-12
Surprisingly slow loading of very simple trees compared to `root_pandas`


daritter:
Running uproot 2.9.3 on python3.6 I see that loading a simple tree (flat, 5 double branches) takes around four times longer than using `root_pandas`.

I have two versions of the input root file: `ntuple-32k.root` with 32kb bucket size (45 buckets per branch) and `ntuple-1M.root` with 1 MB bucket size (2 buckets per branch). In both cases it takes around 70 ms to read the file into a pandas data frame with `root_pandas` but 300ms using `uproot` on my machine.

I attached both files and the ipython log for the benchmark as [ntuple.zip](https://github.com/scikit-hep/uproot/files/2188709/ntuple.zip)

jpivarski:
Thanks! I'll take a look.

jpivarski:
First observation: it's not I/O. WIth similar results between ntuple-32k.root and ntuple-1M.root,

Time to read into arrays is a quarter of a second:

```python
>>> import time
>>> import uproot
>>> t = uproot.open("testfiles/ntuple-32k.root")["variables"]
>>> starttime = time.time(); q = t.arrays(); print(time.time() - starttime)
0.245407104492
```

Time to read into a Pandas DataFrame is more than a second:

```python
>>> starttime = time.time(); q = t.pandas.df(); print(time.time() - starttime)
1.08535289764
```

Reading the arrays and constructing the DataFrame manually is back down to a quarter of a second:

```python
>>> starttime = time.time(); q = pandas.DataFrame(columns=t.allkeys(), data=t.arrays()); print(time.time() - starttime)
0.255214929581
```

I would check with root_numpy, but my root_numpy installation is broken because I updated ROOT recently.

Reading the arrays and manually constructing the DataFrame makes an extra copy of the data— `t.pandas.df()` does something clever to try to preallocate the DataFrame and write the data directly into it. It was supposed to be faster, but it's far slower in this case. As usual, clever is bad.

For now, I'll remove the cleverness until I understand Pandas DataFrame constructors better.

jpivarski:
Done! It's [now available on PyPI](https://pypi.org/project/uproot/2.9.4/). (Time to board a plane. :)

PR #101, opened 2018-07-13
fixes issue #100: simplify construction of DataFrames

jpivarski:
Yes, there's a data copy, but Pandas is much faster this way.


jpivarski:
That Travis failure is a broken Travis server (failed to start after 10 minutes). The identical Python 2.6 test was successful. Also, I'm not going to wait for the AppVeyor tests because nothing Windows-specific is being added in this PR.

PR #143, opened 2018-09-25
Issue133

### LINKED ISSUE #133: Datime timestamps need to be properly formatted
Right now, they're constants. Write an algorithm that converts time from Python's `time` or `datetime` library into this format and properly fill them in everywhere they appear (`TDirectory` and `TKey`; search for "FIXME").

reikdas:
Fixes #133

### DIFF
--- a/uproot/write/TDirectory.py
+++ b/uproot/write/TDirectory.py
@@ -45,7 +45,7 @@ class TDirectory(object):
         self.fSeekDir = fSeekDir
         self.fSeekParent = fSeekParent
         self.fSeekKeys = fSeekKeys
-        self.fDatimeC = uproot.write.util.time()
+        self.fDatimeC = uproot.write.util.datime()
         self.fUUID = b'\x00\x01' + uuid.uuid1().bytes
 
         self.allocationbytes = allocationbytes
@@ -63,7 +63,7 @@ class TDirectory(object):
 
     def update(self):
         fVersion = 5
-        fDatimeM = uproot.write.util.time()
+        fDatimeM = uproot.write.util.datime()
         self.cursor.update_fields(self.sink, self._format1, fVersion, self.fDatimeC, fDatimeM, self.fNbytesKeys, self.fNbytesName, self.fSeekDir, self.fSeekParent, self.fSeekKeys)
 
     def write(self, cursor, sink):
--- a/uproot/write/TDirectory.py
+++ b/uproot/write/TDirectory.py
@@ -31,7 +31,6 @@
 import collections
 import struct
 import uuid
-import datetime
 
 import uproot.write.sink.cursor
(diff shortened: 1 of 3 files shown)


jpivarski:
Hi Reik!

Same issue as before: don't pass dates into the TKey and TDirectory classes; these are internal things that only the class itself should have to deal with. Go back to the original FIXMEs, which were on hidden, local variables, and set them internally. That way, whenever a TKey or TDirectory is written (or updated), it will timestamp itself without the calling code knowing anything about it.

Second part is the DRY (Don't Repeat Yourself) principle: you've written the logic for creating timestamps three times. This should be a single function accessible to TKey and TDirectory or a single superclass (named something like `Timestampable`). For such simple functionality, it should probably be a free-standing function in a library named `uproot/write/util.py`. The test for whether TKey and TDirectory should be subclasses of a `Timestampable` class is this: "are TKey and TDirectory, in some sense, the same kind of object because of this time-stamping ability?" I think the answer to that question is "no," so it should be a free-standing function in a `util.py`.

Thirdly, considering that `DatimeC` means "creation time" and `DatimeM` means "last modified time," `DatimeC` must be a class member defined in the `__init__` constructor and remembered (never updated) thereafter. `DatimeM` must be changed with each `update()`, so it could be a class member or a local variable. Therefore, it should be a local variable. Never make class members (potentially mutable state) unless you have to.

jpivarski:
That failure is irrelevant (an AppVeyor machine couldn't download the LZMA package).

jpivarski:
The ROOT documentation I could find doesn't say whether TKey's `fDatime` is a creation time or a modification time, and since a new TKey is created whenever an object is placed, there probably isn't a distinction. However, the same TKey appears in both the directory and right before the object in the file. The way it's implemented right now, these TKeys could get slightly different `fDatimes`. That's _probably_ a bad idea. I think we should pretend that TKey's `fDatime` is a creation time, define it once in the constructor, and therefore keep it stable. I don't know if letting the TKey time vary would be a bug, but if so, it would be a hard to diagnose bug.

Also, let's generalize the time function a little:

```python
def datime(when=None):
    if when is None:
        when = datetime.datetime.now()
    return (when.year - 1995) << 26 | when.month << 22 | when.day << 17 | when.hour << 12 | when.minute << 6 | when.second
```

Also, I prefer the name `datime` because this puts the timestamp specifically in ROOT's "datime" format, which isn't used anywhere else that I know of. (Including that spelling!) `time` is the name of a Python library and is a very general word.

Other than that, this looks great!

jpivarski:
Looks good! I found out that I could add the license directly through the web form (I'll do that for other trivial patches in the future). Then I forgot to bump the version number, so I did that in master.

Thanks!

PR #147, opened 2018-10-03
changes need to read Geant4 file

jpivarski:


jpivarski:
The thing I was most worried about— backing up and skipping the byte count if it doesn't have the `kByteCountMask` flag set:

```python
def _startcheck(source, cursor):
    start = cursor.index
    cnt, vers = cursor.fields(source, _startcheck._format_cntvers)
    if numpy.int64(cnt) & uproot.const.kByteCountMask:
        cnt = int(numpy.int64(cnt) & ~uproot.const.kByteCountMask)
        return start, cnt + 4, vers
    else:
        cursor.index = start
        vers, = cursor.fields(source, _startcheck._format_cntvers2)
        return start, None, vers
_startcheck._format_cntvers = struct.Struct(">IH")
_startcheck._format_cntvers2 = struct.Struct(">H")
```

is validated [here in ROOT](https://github.com/root-project/root/blob/331efa4c00fefc38980eaaf7b41b8e95fcd1a23b/io/io/src/TBufferFile.cxx#L2786):

```c++
   // no bytecount, backup and read version
   if (!(v.cnt & kByteCountMask)) {
      fBufCur -= sizeof(UInt_t);
      v.cnt = 0;
   }
   if (bcnt) *bcnt = (v.cnt & ~kByteCountMask);
   frombuf(this->fBufCur,&version);
```

So we have confirmation that this is the right thing to do. We've just never seen a file old enough to need it until now.

Issue #148, opened 2018-10-05
Add verbose diagnostic for opening xrootd file


vkuznet:
Jim,
I have two issues with opening XrootD files and it would be helpful if you'll add some verbose output about steps involved in access to XrootD server.

Problem 1: I'm using GPU node (at Cincinatti) and over there when I do uproot.open call it just sits there doing nothing. I don't know what's wrong and I can't figure out where to look at. The file I'm trying to open is
root://cms-xrd-global.cern.ch//store/data/Run2018C/Tau/NANOAOD/14Sep2018_ver3-v1/60000/6FA4CC7C-8982-DE4C-BEED-C90413312B35.root

Problem 2: on my mac I can't open this file either, but I get some output which I don't know how to interpret:
```
>>> data=uproot.open(fin)
Traceback (most recent call last):
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/cachetools/cache.py", line 51, in __getitem__
    return self.__data[key]
KeyError: 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/uproot/source/chunked.py", line 84, in data
    chunk = self.cache[chunkindex]
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/uproot/cache.py", line 79, in __getitem__
    return self._cache[where]
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/cachetools/lru.py", line 16, in __getitem__
    value = cache_getitem(self, key)
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/cachetools/cache.py", line 53, in __getitem__
    return self.__missing__(key)
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/cachetools/cache.py", line 80, in __missing__
    value = self.__missing(key)
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/cachetools/cache.py", line 110, in __missing
    raise KeyError(key)
KeyError: 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/envs/fastai-cpu/lib/python3.6/site-packages/uproot/rootio.py", line 75, in open
    return xrootd(path, xrootdsource)
[...]

PR #149, opened 2018-10-05
Trigger Binder to build image on pushes to master

matthewfeickert:
Use selenium to trigger Binder to have repo2docker start the build of the image anytime a new push to master occurs. This ensures that there will always be a recently built Binder example so that if someone clicks on the launch Binder badge they don't have to wait for it to build. selenium is used instead of just `python -m webbrowser` as the browser needs to be open on the URL long enough for Binder to trigger.


matthewfeickert:
@jpivarski Since your Travis setup is a bit different then the one we use for pyhf, so I need to figure out how to arrange things properly to work with it, but I think the idea for the PR is clear.

jpivarski:
As long as this doesn't interfere with normal testing and deployment, I'll take it, thanks!

Could you add the [version-3-features.ipynb](https://github.com/scikit-hep/uproot/blob/master/binder/version-3-features.ipynb) to the list? (I see a list under jobs/scripts, which I assume we have to keep updated with each new notebook.)

It would be even cooler if it evaluated the notebooks, updated the outputs, and triggered a Travis failure if any exceptions were raised. I've been testing version-3-features.ipynb manually to make sure that the updates I'm making to awkward-array don't break uproot because I haven't converted the notebook into tests yet.  :)

matthewfeickert:
> It would be even cooler if it evaluated the notebooks, updated the outputs, and triggered a Travis failure if any exceptions were raised.

Oh yeah, we use [papermill](https://papermill.readthedocs.io/en/latest/) to do this. I can add this in another PR.

jpivarski:
Okay, if it's not working yet, let me know when it's done. I don't know much about this part of Travis and Binder (let alone "selenium").

jpivarski:
If it's not too much work and you already know how to do it, yes please! I just checked tutorial.ipynb and version-3-features.ipynb, and they're both executed from top to bottom, no exceptions expected. Whenever I wanted to show that a particular call is supposed to raise an exception, I do it in a try-except block. If Papermill would make Travis turn red and not deploy broken notebooks to Binder on exceptions, that would be a great way to make sure that the tutorials are up to date. (Readthedocs does similar for references.)

jpivarski:
> I might have to rewrite a large chunk of the YAML, so please feel free to give lots of veto feedback on that.

Okay. It makes me a little nervous because these are the parts of the process that I don't fully understand. I'll need to see it testing the full testing matrix and do one trivial deployment to make sure that it can still deploy. There's a little bit of tricky configuration to make sure that only one of the test cases attempts to deploy, so that it's not a race condition and the non-first ones don't fail because the artifact is already deployed.

matthewfeickert:
@jpivarski This now triggers Binder to build the image for
- the entire repo
- [binder/tutorial.ipynb](https://github.com/scikit-hep/uproot/blob/master/binder/tutorial.ipynb)
- [binder/version-3-features.ipynb](https://github.com/scikit-hep/uproot/blob/master/binder/version-3-features.ipynb)

on each **push to master**, so that they only trigger once there is something new to update.

Likewise, **deployment only happens on pushes to master** now. I'm not sure if you want this or not, but this is what we do for pyhf as there doesn't seem to be any reason to deploy to PyPI if you aren't going to be doing a new release, and that would happen when you make a tag on master.

So the test matrix is left untouched and now these new stages will only appear on direct pushes to master or when a PR is merged in.

If you want to see what this looks like with all the stages then I can do a commit that has them run on any push to master and then if you're satisfied we can revert that before squashing and merging. Thoughts?

matthewfeickert:
You'll notice that I've also turned on caching of pip, which drops the total testing time by about 2 minutes.

jpivarski:
What about `if: (tag =~ v[0-9\.]*)`? Does that mean that it will only deploy if version tags start with a `v`? Mine don't: [see tags](https://github.com/scikit-hep/uproot/tags). The tag names are exactly equal to the version strings, which are prescribed by one of the PEPs. I don't make tags other than for deploying releases. (I use branches for saving points in history because they're often moving points in history: e.g. "version 1.x with bug-fixes.")

I would only want to deploy on master, that's true, but I also only make tags/releases on master, after a successful test. The test right before the deployment is superfluous, and it looks like you removed it, which would be good.

If this travis.yml stops working someday (due to changes in Travis— it's happened), would I be able to get in touch with you for how to fix it? Otherwise, I'd have to start pulling apart features to return to something minimal that works.

Do I understand correctly the sequence:
   1. PRs work like they used to: each PR commit runs tests and only tests.
   2. When a PR is merged into master, that runs the tests and also updates Binder.
   3. A raw commit to master (I do it sometimes for very simple things; less and less as uproot gets complex) also runs the tests and updates Binder.
   4. A new release >>> runs tests? does not run tests? <<< and then deploys to PyPI.

I'll have a real update to test the system on Monday: I'm going to partially fix #148 by adding a `timeout` option to `uproot.open` with XRootD. That will go through the whole PR-merge-release sequence.

matthewfeickert:
> Does that mean that it will only deploy if version tags start with a `v`?

Ah yes, my bad. You make a good point, and thinking about this I've actually gone way outside the scope of what this PR should be. Let me reorder this in another commit that only implements the Binder trigger and then in another PR I can change the CI file more and we can discuss there.

> If this travis.yml stops working someday (due to changes in Travis— it's happened), would I be able to get in touch with you for how to fix it?

Yes, of course! However, I think that any changes I'm going to implement will be really quite small, so they shouldn't be too bad to understand I think. :)

> Do I understand correctly the sequence:
> 1. PRs work like they used to: each PR commit runs tests and only tests.
> 2. When a PR is merged into master, that runs the tests and also updates Binder.
> 3. A raw commit to master (I do it sometimes for very simple things; less and less as uproot gets complex) also runs the tests and updates Binder.

Yes.

> 4. A new release >>> runs tests? does not run tests? <<< and then deploys to PyPI.

As a new release is simply just a commit to some branch (and probably on master) then the CI will run normally and run all the tests and then deploy if the tests succeed.

matthewfeickert:
@jpivarski Okay, I've reverted the deployment changes (those can be in another PR so that this one stays in scope to be just about Binder). So this is ready for another review.

My proposed commit log message for the squashed commit for this PR is the following:
>* Use selenium to trigger Binder to have repo2docker start the build of the image anytime a new push to master occurs. This ensures that there will always be a recently built Binder example so that if someone clicks on the launch Binder badge they don't have to wait for it to build. selenium is used instead of just `python -m webbrowser` as the browser needs to be open on the URL long enough for Binder to trigger.
>* Turn on caching of pip in Travis for faster builds
>* Use Travis CI stages to separate the testing stage from the Binder triggering stage

matthewfeickert:
> I guess I wonder what's in that "pip cache" and if it might prevent the latest awkward-array or uproot-methods from being tested.

We (the pyhf dev team) wondered about this as well, but it turns out that if the install procedure would install a different version it will go ahead and do so. The cache just exists to speed up the installation of versions that aren't new.

Issue #151, opened 2018-10-06
Support IPython completion 


henryiii:
IPython supports key completion in dictionary like objects. In order to support it, a list of possible key completions should be returned from `_ipython_key_completions_()`. If you manually hack this in a notebook:

```python
d.__class__._ipython_key_completions_ = lambda x: [x.decode() for x in x.keys()]
```

Then type:

```python
d[
```

Then press tab, you'll get `'HitStrips;1'` as a possible completion (in my case, `d` is a `TDirectory`). The ";1" could also be stripped on directories, as it's not needed as a key.

It would be nice if all dictionary-like classes supported this method in uproot.

Edit: My usage of decode in the hack is Python 3.

jpivarski:
Fixed by #152 in uproot version 3.2.1.

PR #152, opened 2018-10-06
Adding support for IPython key completions

henryiii:
Addressing #151 .

Note that completion looks significantly more useful if you press tab before you type a quote - then the key completions are at the top because it sorts the completions, and the quote `'` causes the key completions to be at the top. Not sure why Jupyter assumes any other sort of completion inside a string for a key...

Only tested locally on Python 3, will use CI to test Python 2.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -243,7 +243,7 @@ In addition, `ROOTDirectory <http://uproot.readthedocs.io/en/latest/root-io.html
 
     >>> for n, x in file.allclasses():
     ...     print(repr(n), "\t", x)
-    ... 
+    ...
     b'one;1'          <class 'uproot.rootio.ROOTDirectory'>
     b'one/two;1'      <class 'uproot.rootio.ROOTDirectory'>
     b'one/two/tree;1' <class 'uproot.rootio.TTree'>
@@ -461,7 +461,7 @@ The **outputtype** parameter lets you specify the container for your arrays. By 
 
     >>> for x, y, z in tree.iterate(["Float64", "Str", "ArrayInt32"]):
     ...     print(x, y, z)
-    ... 
+    ...
     ArrayInt32 Str Float64
 
 A ``for`` loop over a ``dict`` just iterates over the names. We've read in three arrays, thrown away the arrays, and returned the names. In this case, we really wanted a tuple, which drops the names (normally needed for context), but preserves the order and unpacks into a given set of variables:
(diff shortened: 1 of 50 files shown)


henryiii:
PS: I didn't add this method for the custom OrderedDict, since it is left over from Python 2.6 support. I wasn't sure if the file `uproot/write/TFile.py` should have a redirect method for this like the others, depends on if it is user-facing. For now, I've added it.

jpivarski:
Thanks! I had starred your issue to deal with on Monday— it sounds like a great idea. It could apply to TTrees and TBranches as much as TDirectories (and write TFiles). The `;N` is only ever needed if the key name without it is not unique— it doesn't actually have anything to do with directory vs other kind of object.

To see this, consider writing two histograms to a file with the same name. You'd expect the second to overwrite the first, but both get written with different cycle numbers. That's rather unpythonic, but it's a bridge to how ROOT works. Only if cycle numbers are included in the key, it's one key is one object.

henryiii:
I've removed the filtering on TDirectory, just in case there is a `;2` or more. A smart filter could be added, but for completion, I think it's fine to compete with the `;1`, it's both correct and easy to erase.

The current patch does apply to the other cases as well, anything that has a `def keys`, in fact. Select "ignore whitespace for now" on the github diff settings; apparently uproot is a bit messy on ending whitespace. Having your editor filter ending whitespace is not a bad idea, IMHO. ;) I can remove the whitespace changes if you want.

jpivarski:
Oh, you're right: there it is in uproot/tree.py. I don't mind the whitespace changes at all— I remove them when I discover them manually (through a nervous twitch where I control-E at the ends of lines). I suppose it should be something for the editor to do automatically, though.

The test failures were Travis's fault— I reloaded them and they're fine.

jpivarski:
Could you change the version number in uproot/version.py to 3.2.1 so that when I merge this it will be ready to be a new release?

henryiii:
Sure. Is the there a changelog somewhere, or just the GitHub releases?

henryiii:
Bumped version, and removed all EOL whitespace.

On macOS, that's:
```bash
git ls-files '*.h' | xargs sed -i '' 's/[[:space:]]*$//'
```

per file type. (on Linux, you'd drop the `''` after `-i`).

jpivarski:
Thanks! I know of a way to edit through the website for non-pull requests, now I just need to learn to do that for pull requests (because we always need a version bump). I don't like doing that on master because it triggers more CI machinery (and using "[skip ci]" would then add that message to the last commit before the release, causing the release to not do CI, which is bad because that's how we deploy!).

Changing every single file in the repository is a bit terrifying, but I see nothing wrong with this particular change. The tests were successful, so now I'll merge. Once that goes through, I'll deploy a new version to PyPI and we'll see a complete test of #149.

jpivarski:
> Sure. Is the there a changelog somewhere, or just the GitHub releases?

There's just the GitHub releases, which taken together has the contents of what I would put in a CHANGELOG. Do you have a reason why a single file is preferable?

If so, I could start that discipline by gathering release messages and then keeping them in sync from here on out.

henryiii:
For a CHANGELOG in the source, it's nice, but releases work too. As long as there's a log somewhere...

jpivarski:
The full chain seems to have worked! This is now in PyPI. I'll start on an AUTHORS file on Monday morning.

PR #154, opened 2018-10-08
[WIP] Add tests for Jupyter Notebooks

### LINKED ISSUE #150: Test Jupyter Notebooks in papermill
The Jupyter notebooks should all be run in the CI with papermill to make sure that active development doesn't break documentation.

`pyhf` does this, so for reference see the [notebook tests that we have for `pyhf`](https://github.com/diana-hep/pyhf/blob/master/tests/test_notebooks.py).

matthewfeickert:
Resolves #150


jpivarski:
Cool! I'll check in on this again tomorrow morning.

You know, the `wget` commands are unnecessary and I've been meaning to get rid of them in favor of symbolic links: they download files that can be found in `../tests/samples/". You don't need to do that replacement, but if it simplifies things while you're working on the postBuild (maybe eliminating it altogether), I wanted to let you know about this fact.

matthewfeickert:
Thanks for the info. This is going to need a little bit of extra work as

```
python -m unittest
```

passes on my local machine, but it is failing on Travis. However, `unittest` is just giving `ERROR` as the only feedback. I'll have to check to see if `unittest` can be made to give stacktraces like `pytest` does.

matthewfeickert:
Hm, [it looks like modern Python and NumPy > 1.15 are needed here](https://travis-ci.org/scikit-hep/uproot/builds/438433882), so I'll probably break this out into its own CI stage. The only trick now is how to have `unittest` ignore a test file. In `pytest` this would just be

```
pytest --ignore tests/test_notebooks.py
```

but I'm not quite sure how you do this in `unittest` from the CLI.

jpivarski:
I frequently do the opposite of that (run only one test from the command line) with `-s`.

matthewfeickert:
> I frequently do the opposite of that (run only one test from the command line) with `-s`.

Hm, maybe I'm misunderstanding what you mean. My thought was that I would figure out how to have `unittest` ignore `tests/test_notebooks.py` during the normal run of the tests in the CI and then in a separate Travis stage run only the notebook tests with

```
python -m unittest tests/test_notebooks.py
```

Is there a better workflow for this?

jpivarski:
My first comment wasn't useful since it was the opposite of what we want here: `-s` will run one test, rather than all-except-one.

Another solution to control it from the commandline is to use `-k` to select tests: it selects them by substring and we could use a different substring for the Jupyter tests. It is the case that 100% of the tests in the current test suite start with the substring "test_". This was to satisfy a Python 2.6 constraint, which is no longer relevant but the naming convention is still there. The new test could start with a different substring, like "jupyter_" and then they're disjoint: `python setup.py test -k test_` and `python setup.py test -k jupyter_`.

However, I like your solution better. By decorating the Jupyter test with version constraints, it will be executed in the normal test run, but I don't think that's a bad thing. Also, the decorator says exactly under what conditions the test will run— it's not controlled by a commandline switch far away.

This does make the later-version end of the test suite more heavyweight. Since we do get some parallelism (4 run at a time), reversing the order so that the longer-running tests start first would cut down on the total running time. I've been meaning to do this for a while with the PyPy tests, which take three times as long as all the others. But now the same will be true of Pythons 3.5 and 3.6.

jpivarski:
I'm going to put issue148 in as version 3.2.2 because this is still having test failures. (Perhaps it's ignoring or misinterpreting the Python version directive?) When this is complete, it will have to merge in the uproot/version.py update.

jpivarski:
@matthewfeickert Is this and issue #150 not going to be completed? If so, that's okay; I'm just cleaning up things like this.

matthewfeickert:
Hi @jpivarski. It will be, but given that it is going to take me some time (trying to focus on finishing my Ph.D. and finding a job atm). Maybe we can close them for the time being and then when I get time I can reopen them?

jpivarski:
That sounds good. You have your branch, which you can use to reopen, but by the time you get around to doing that, you'll probably want to start fresh from the top of master. (I'm asking about this one now because it might be hard to get up to date.)

Meanwhile, good luck on your degree!

matthewfeickert:
> You have your branch, which you can use to reopen, but by the time you get around to doing that, you'll probably want to start fresh from the top of master

Oh yeah, I'm definitely going to rebase against `origin/master`. In general if I can't do a ff merge then I'm going to want to take time to interactively rebase a PR until I can.

Issue #156, opened 2018-10-09
Filtering pandas df


henryiii:
Currently, there's only one option for reading pandas DataFrames from TTrees that have JaggedArrays: `flatten=True`. However, another use case would be to have a simple way to filter JaggedArrays out.

Currently, this looks like:

```python
good_keys = [key for key in t.keys() if not isinstance(t[key].interpretation, uproot.asjagged)]
t.pandas.df(good_keys, ...)
```

This might be useful in the function call instead. Ideas: `filter=True`, `filter_jagged=True` ,`flatten=None`, `flatten='filter'`, or `jagged='filter'` (`jagged='flatten'` or `multi-index` being another policy) are some ideas.

(This is because I'm dealing with everything-but-the-kitchen-sink NTuples...)

jpivarski:
Actually, this is broader than the DataFrame interface: it affects the set of branches that are loaded if none are specified. The `branches` parameter [already takes a number of options](https://uproot.readthedocs.io/en/latest/ttree-handling.html#arrays), one of which is a function from branch (object) to interpretation or None. You can right now do something like

```python
t.pandas.df(lambda b: b.interpretation if not isinstance(b.interpretation, uproot.asjagged) else None)
```

though this is far from obvious. Maybe there should be a shorthand object that the `branches` parameter understands to define such policies in a more user-friendly way? One I can think of: suppose any Interpretation class object like `asdtype` is taken as

```python
lambda b: b.interpretation if isinstance(b.interpretation, asdtype) else None
```

But then, you want the _inverse_ of `asjagged`.

henryiii:
I did notice that it took a function, but I haven't taught my students lambda functions and I have taught them comprehensions, so that's why I structured my example that way.

However, `.df` already has a parameter that is designed for jagged arrays. What about having flatten take True (Current behavior), False (new: filter out jagged arrays), and None (the default, current behavior, crash)

henryiii:
Sorry, the default is jagged=True, and the crash is not because it's not trying to flatten, it's because:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-13-2689fd22c539> in <module>()
----> 1 df = t.pandas.df(entrystop=100_000)

~/.local/lib/python3.6/site-packages/uproot/_connect/to_pandas.py in df(self, branches, namedecode, entrystart, entrystop, flatten, cache, basketcache, keycache, executor, blocking)
[...]

PR #157, opened 2018-10-10
Adding test and support for None for flatten

henryiii:
This addresses #156 , it makes `flatten=None` a possible option when filtering; unlike `False`, this simply removes anything that would need flattening, rather than making them lists. It also adds an *expected failing* test for the current broken `flatten=True` behavior.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ matrix:
       env: NUMPY="numpy==1.13.1"
 
 install:
+  - env
+  - echo $TRAVIS_PYTHON_VERSION
   - pip install --upgrade setuptools_scm
   - pip install $NUMPY
   - python -c 'import numpy; print(numpy.__version__)'
@@ -37,6 +39,7 @@ install:
   - python -c 'import awkward; print(awkward.__version__)'
   - pip install "uproot-methods>=0.2.0"
   - python -c 'import uproot_methods; print(uproot_methods.__version__)'
+  - if [[ $TRAVIS_PYTHON_VERSION != pypy* ]] ; then pip install pandas ; fi
 
 addons:
   apt:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -47,4 +47,5 @@ build_script:
   - "pip install %NUMPY%"
   - "pip install uproot-methods>=0.2.0"
   - "pip install -i https://pypi.anaconda.org/carlkl/simple backports.lzma"
+  - "pip install pandas"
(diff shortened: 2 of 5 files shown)


henryiii:
Tests are passing now. Added pandas testing, but not on PyPy. (The `=True` bug is still present, I think you can fix that more easily than I).

jpivarski:
I plan to check this, correct the None case, and migrate to pytest tomorrow.

henryiii:
Since the default case is `True` and is currently broken, you might want to think about setting "False" as the default when you fix it - since no one can be using the default at the moment. That would be more consistent with the rest of uproot, and is a less surprising default, I think. Just a thought.

jpivarski:
I'll fix it before merging this PR. The only exceptions I want the user to see are ones that say how their inputs were wrong (or external things, like filesystem access, etc.).

As for consistency, I actively decided that `flatten` would be `False` for all array-fetching methods except `TTree.pandas.df`. This one is different because it's a connector to an external library, and the usual thing you'll want to do in that library is to flatten the columns, putting the structure into the index.

Actually, there are two ways to fill a DataFrame. You can use the normal array-fetching methods (`TTree.arrays`, etc.) with `outputtype=pandas.DataFrame` or the connector method. In the former, you get the normal defaults because you're just saying that you want a different output type; in the later, you're primarily thinking about Pandas and what's normal in that world. That's the reason for the different defaults.

In fact, the connector method is just a repackaging of the normal method with different defaults. That's all it does. (As you've seen.) Is that a terrible idea? Should consistency win over my guess at user expectations? (Although it's probably too late to change now— it's been public API for a while.)

henryiii:
However, given that the public API is currently broken, it gives us some leeway to change it. I don't think I'd recommend getting rid of it, since it's more discoverable than `.array(outputtype=pandas.DataFrame)`, but having matching defaults might be better if this is the only one that's different. What defaults are different, besides the flatten?

Also, flatten=True in my case (300+ regular columns and a small handful of jagged columns) would probably waste much more memory than flatten=False, and I don't think most users expect multi-index DataFrames.

jpivarski:
I'm closing this here to continue on #160 (this PR + my changes).

henryiii:
You can always push to my branch, by default you have access to when someone makes a PR. But that's fine. :)

jpivarski:
Darn; I wanted to do that but the text output from my git command suggested this way, so I thought I couldn't push to it. I'll do more Googling and get it right next time.

PR #160, opened 2018-10-11
Henryiii filterpandas

jpivarski:
Continuation of PR #157 with Jim's bug-fixes.

### DIFF
--- a/uproot/tree.py
+++ b/uproot/tree.py
@@ -429,7 +429,7 @@ class TTreeMethods(object):
         entrystart, entrystop = self._normalize_entrystartstop(entrystart, entrystop)
 
         # start the job of filling the arrays
-        futures = [(branch.name if namedecode is None else branch.name.decode(namedecode), interpretation, branch.array(interpretation=interpretation, entrystart=entrystart, entrystop=entrystop, flatten=(flatten and not ispandas), cache=cache, basketcache=basketcache, keycache=keycache, executor=executor, blocking=False)) for branch, interpretation in branches]
+        futures = [(branch.name if namedecode is None else branch.name.decode(namedecode), interpretation, branch.array(interpretation=interpretation, entrystart=entrystart, entrystop=entrystop, flatten=(flatten and not ispandas), cache=cache, basketcache=basketcache, keycache=keycache, executor=executor, blocking=False)) for branch, interpretation in branches if not (isinstance(interpretation, asjagged) and flatten is None) ]
 
         # make functions that wait for the filling job to be done and return the right outputtype
         if outputtype == namedtuple:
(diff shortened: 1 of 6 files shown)


jpivarski:
> However, given that the public API is currently broken, it gives us some leeway to change it. I don't think I'd recommend getting rid of it, since it's more discoverable than .array(outputtype=pandas.DataFrame), but having matching defaults might be better if this is the only one that's different. What defaults are different, besides the flatten?

At best, you can say that the public API is misguided. I thought about it, concluded that this way would be least confusing for users, implemented it, and used this used these defaults in code examples I gave in talks. I might have made a wrong choice, but "broken" should be reserved for unintentional mistakes, not intentional ones.

In doing so, I decided on a general rule that connectors `TTree.nameoflib.func(...)` (where `nameoflib` is a famous library that the user cares about) can have different defaults than raw array-fetching functions `TTree.func(...)`. So far, `flatten` is the only one that yet takes advantage of that freedom.

> Also, flatten=True in my case (300+ regular columns and a small handful of jagged columns) would probably waste much more memory than flatten=False, and I don't think most users expect multi-index DataFrames.

Your case has a much higher flat/jagged ratio than the cases I've considered so far. CMS NanoAOD only has more flat branches than jagged branches because each trigger is a separate branch of booleans. Excluding those, maybe 90% of what's left is jagged.

Currently, `flatten=False` means losing the Numpyness of the branch by iterating over it, turning the elements into Python lists. That's a performance issue in memory and load time. Then you have a DataFrame with a `dtype=object` column, which you can't do anything with except Python loops. Soon, though, I'll be adding Pandas extension types for jagged arrays that would make `flatten=False` a lot more performant and useful.

In those talks I referred to, I advocated for more use of Pandas MultiIndex, since it's the Pandas-native way of dealing with this sort of structure. Maybe most users aren't familiar with it now, but I hope that will change.

uproot/tree.py
```diff
@@ -423,6 +423,9 @@ def array(self, branch, interpretation=None, entrystart=None, entrystop=None, fl
 
     def arrays(self, branches=None, outputtype=dict, namedecode=None, entrystart=None, entrystop=None, flatten=False, cache=None, basketcache=None, keycache=None, executor=None, blocking=True):
         branches = list(self._normalize_branches(branches))
+        if flatten is None:
+            branches = [(branch, interpretation) for branch, interpretation in branches if not isinstance(interpretation, asjagged)]
+            flatten = False
```
henryiii: None evaluates to False, so line 428 is not required.

jpivarski: Yeah, but as I said before, I don't rely on that in my internal code. When I say `if x`, I manually ensure that `x` is boolean. This is a style choice that may or may not agree with a PEP, but I'm sticking to it (and enforcing it in my projects) because I think it's safer.

henryiii: That's fine. Thanks for the fix!

Issue #158, opened 2018-10-10
Migrate from unittest to pytest


jpivarski:
It's what all the cool kids are using.

jpivarski:
Done. Not a new version because this only affects testing.

PR #163, opened 2018-10-15
switch from unittest to pytest

jpivarski:
The first commit uses pytest's unittest compatibility to simply switch over: is there anything else to do?

Addresses #158.


jpivarski:
@henryiii (in #156) and @matthewfeickert (in #150 and #154), you both recommended for me to start using pytest. I looked into it this morning and this tiny PR seems to do it: now when I run `python setup.py test`, it runs pytest instead of unittest. So far, the continuous integration is also running pytest with no troubles.

Since pytest doesn't have any of its own classes, my unittest-based classes still work. So while this may be a formal migration, it's not very pytesthonic. Is there something else I ought to do to provide the benefits you were looking for? I don't plan to migrate _back_ to unittest, so I'd happily accept pytest decorators that don't work in unittest, and you get pytest's commandline control over which tests run, as well as its output format. What more would be needed?

Maybe translating names like `assertRaises` to `raises`? I used very little unittest functionality, just its wrapped assert methods, so that could even be a textual search-and-replace. Is there any value in doing so? If so, from what to what?

Thanks!

henryiii:
Plumbum also has detailed PyTest tests, including an example of a `conftest.py` file and custom fixtures. It uses skipping decorators extensively, as well. <https://github.com/tomerfiliba/plumbum/tree/master/tests>

matthewfeickert:
>  So while this may be a formal migration, it's not very pytesthonic.

I'll just echo everything that @henryiii has said above. If you want to switch over now you could just merge this PR and then in another we can make the tests be properly `pytest`.

There's nothing wrong with what you're doing to run the tests at the moment, but it might be a bit more clear if in `.travis.yml` instead of having

```
python setup.py test
```

you instead had something along the lines of

```YAML
install:
# stuff
  - pip install -e .

# more stuff

script:
  pytest
```

jpivarski:
I decided to convert them anyway, to set a good example for new code. I used very few features of unittest (and now use very few features of pytest).

With Emacs keyboard macros, I could convert `assertTrue` tests with zero errors (because of sexp keystrokes). Unfortunately, there's not a good keystroke for jumping to the correct `,` to convert to `==` in `assertEqual` because Python is not as regular as LISP. I was just careful. The fact that there's the same number of tests before and after conversion plus they all pass convinces me that they were converted correctly. (I can see how I might have accidentally lost one, but I can't see how I could have accidentally gained one. I hadn't used any bare `asserts` before.)

Now the word "unittest" doesn't appear anywhere in the codebase, so it's done.

matthewfeickert:
It seems that most of these tests aren't actually using features of a pytest class, and so for clarify they might benefit from just existing as their own tests (much as they are now) in the files. They could probably benefit more from adding some [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html).

I can help with those in another PR later on if you're interested.

jpivarski:
If it would help developers, then I'd approve the PR. At the moment, I can't think of any features that a testing suite needs other than discoverability, coverage, and good error reporting. (I'm sure coverage could be improved, but that's not a pytest feature.) Let me know what you have in mind. Thanks!

matthewfeickert:
Fixtures can be really nice for properly modularizing tests. `pytest` has some other features that can lead to improved robustness of tests as well (check out a nice example that @betatim and @lukasheinrich [give here](https://twitter.com/betatim/status/1052117903112323072)) but I can elaborate more on that in another Issue (or on slack).

Issue #139, opened 2018-09-20
Unit tests for writing!


jpivarski:
The writing feature doesn't "fully" exist until there are unit tests. There's a stub for adding unit tests, such as the ones we've been running interactively. I'll add some, you add some, and I'll find a way to put PyROOT in the Travis tests for automatically verifying that ROOT can read what we write.

reikdas:
root_pandas seems to have a way to put PyROOT on Travis.  https://github.com/scikit-hep/root_pandas/blob/master/ci/install.sh

jpivarski:
We can look at that, but I've also been talking with @aperloff about how he's set up CVMFS on Travis for accessing ROOT (and many other things, like CMSSW). We should probably start by using his Docker image. If that works, he knows how to customize these images and perhaps we can get one that's more streamlined to our case. Right now, Travis spends a lot more time installing software than performing tests, especially in the PyPy case, but it wouldn't have to if these things were in the image.

There's also a divide-and-conquer aspect to this: we can write tests that only run when ROOT is installed independently of getting ROOT on Travis. That's something that I can check just by checking it out in git.

By the way, I saw your two pull requests and I'll be giving them a close look and probably merging then on Monday. Thanks!

Justin-Tan:
Hi guys, any idea when writing functionality will be implemented? Great package by the way, the ROOT-independency makes data analysis a lot more flexible.

reikdas:
@Justin-Tan We can already write ROOT files with TH1s (and TObjStrings). We plan to add support for TH2, TH3 and TProfile soon.

jpivarski:
I said it doesn't "fully" exist in that sense that if it isn't tested, it's not trustworthy. But you can currently write files with uproot. Instead of opening a file with `uproot.open`, open it with `uproot.recreate` (create a file or overwrite it if it doesn't exist). Then this dict-like object can be used for assignment:

```python
file = uproot.recreate("filename.root")
file["histogram_name"] = histogram
```

Currently, only histograms can be written, though those histograms can be in a variety of formats (e.g. what you get from extracting a histogram from ROOT, what you get from `numpy.histogram`, a Pandas DataFrame with an interval index, Physt histograms, etc.).

Writing TTrees is a separate topic and could be 6 months away, if that's what you're asking.

reikdas:
Should we close this? 
We have tests for all (working) features that were implemented. (The other tests will come in with the fixes for the pending issues.)

jpivarski:
[...]

PR #183, opened 2018-11-13
Add ROOT to Travis CI

chrisburr:
Following up on #139.

I've been meaning to create a conda package for ROOT since the [new compilers tools](https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html) were added in Anaconda 5. This effort has been waiting on conda-forge  migrating to the new compilers ([current status](https://conda-forge.org/status/)). It has taken a while but is now progressing nicely so I've started building packages on [CERN's GitLab instance](https://gitlab.cern.ch/cburr/root-conda#root-in-conda). I will try to add these these packages to conda-forge once the migration is complete.

This PR adds a new test to the build matrix which uses Python 3.7 to run a small test which is skipped if ROOT is unavailable. I also have packages for Python 2.7 and Python 3.6 so these can be added as well. Alternatively, feel free to close this PR if you'd rather use a different method of getting ROOT into the tests.


chrisburr:
Just to add I'm now using this for `root_pandas` (https://github.com/scikit-hep/root_pandas/pull/77)  and I'll make a PR to `root_numpy` soon.

jpivarski:
If you've actually developed a new conda package for ROOT, that's awesome and goes well beyond testing in uproot! (And root_pandas.) Also, this would be preferred over the method I was considering: CVMFS through a custom Travis image.

You should get in touch with Guilherme (https://root.cern.ch/team), who was working on exactly this, the last I heard. Does your solution only work for Python 3.7? If so, then it won't be available to most physicists right away, but it's important to know that there's a streamlined way to do it in the future, because it means we shouldn't spend too much effort on temporary solutions.

I tried to load your ROOT in a test environment:

```
conda create -n testroot -c chrisburr python=3.7 root
```

and perhaps I don't know how to do it because it gave me a list of `PackagesNotFoundError` (root, davix, tbb...).

As for this pull request: could you put this 3.7 test in line with the others, as part of the testing matrix that includes the three versions of Numpy? It shows up in Travis simply as "Python" (no version, no environment variable declaring the Numpy version: https://travis-ci.org/scikit-hep/uproot/builds/454564161?utm_source=github_status&utm_medium=notification). Also, I don't think it's working. I don't see a confirmation that the new test isn't skipped and there's an `atexit` error:

```
Error in atexit._run_exitfuncs:
  File "/home/travis/miniconda/envs/testenv/lib/python3.7/site-packages/ROOT.py", line 777, in cleanup
    facade = sys.modules[ __name__ ]
KeyError: 'ROOT'
```

The tests pass, but possibly without checking what we want to check.

chrisburr:
> You should get in touch with Guilherme.

I'm already planning to!

> Does your solution only work for Python 3.7?

This works for all Python versions supported by conda-forge (2.7/3.6/3.7).

> I tried to load your ROOT in a test environment:

The package still isn't complete due to dependencies which haven't been ported yet in conda-forge but I'll . I suspect graphics and OpenGL are broken but I think everything will work in batch mode. If you managed to break it let me know.

For installing you also need to specify the `gcc7` label in the `conda-forge` channel:

```bash
conda create -n testroot -c conda-forge/label/gcc7 -c chrisburr python=3.7 root
```

> As for this pull request: could you put this 3.7 test in line with the others, as part of the testing matrix that includes the three versions of Numpy?

Yeah I only did it like this to give a minimal PR to start with. I happy to extend this PR to replace the Python (2.7/3.6/3.7) part of the build matrix to use Python from conda-forge instead of Travis.

> I don't see a confirmation that the new test isn't skipped and there's an atexit error:

I'm fairly sure this is a harmless warning caused by ROOT getting confused by `pytest.importorskip` but I'm looking into it.

chrisburr:
[Yup, the test is able to fail.](https://travis-ci.org/chrisburr/uproot/jobs/454589311)

Also this also includes an `pyxrootd` package so `XRootD` tests are now possible (`root://eospublic.cern.ch/` is a good source of test datasets).

jpivarski:
I'm happy if you're happy! If conda doesn't support old versions of Python 3, then we can get rid of them to reduce the size of the matrix. (I only haven't put in 3.7 yet because I think it's been getting out of hand— it takes a long time to wait for a PR to be validated and propagate to a new release.)

However, I don't know how well-used the early Python 3's are. I'm pretty sure that 3.3 is irrelevant. My old laptop was based on Ubuntu 14.04 LTS, whose python3 was 3.4. Maybe that's relevant? I know that I've caught bugs in 3.5, based on gotchas like the fact that the json module didn't accept bytes for a while.

My intention is to accept this PR, though I'm going to be out of touch for a few days due to travel.

Thanks!

jpivarski:
If pyxrootd is available from conda, I'll change the instructions on the main README to include a conda install line. Most of the people I personally know using uproot use it with XRootD.

Semi-related: I've been thinking about putting uproot in conda for a while now, but I just haven't moved to figure out how to do that. Maybe the right thing to do is to make a `scikit-hep` channel, so that it looks official and can include the other packages from this project. Then the front-page instructions would be

```
conda install -c scikit-hep uproot pyxrootd

                  - or -

pip install uproot    # no XRootD
```

I wouldn't make XRootD an explicit dependency, even in conda, but having uproot and its dependencies in conda would make it easy to install them in one go.

Is that something we can do easily?

chrisburr:
I'll have a think about which Python versions to test and get back to you.

> If pyxrootd is available from conda, I'll change the instructions on the main README to include a conda install line. Most of the people I personally know using uproot use it with XRootD.

I wouldn't add them to the README until we pick a channel but you can get XRootD and its Python bindings using:

```bash
conda create -n testxrootd -c conda-forge/label/gcc7 -c chrisburr python=[2.7|3.6|3.7] xrootd
```

> Maybe the right thing to do is to make a scikit-hep channel, so that it looks official and can include the other packages from this project.

I'm split here between:
 - using a dedicated channel like which is built on top of conda-forge like [bioconda](https://bioconda.github.io/)
 - pushing everything upstream to conda-forge once the compiler migration is complete and creating deployment scripts which automatically update it

I think it's a question of scope, can we imagine the channel growing to contain many packages, MC generators and other tools? Or do we just want to put a handful of scikit-hep packages on conda-forge?

If we go for a dedicated channel, we need to consider what to do for storage and build machines as I believe both conda-forge and bioconda rely upon agreements they have with CircleCI and Anaconda, Inc. Anaconda.org only provides 3GB of space but we could work around that by hosting it elsewhere and having a less nice channel name such as `https://scikit-hep.s3.cern.ch`. For building Linux support is fairly easy but providing macOS or Windows builds is trickier.

jpivarski:
You're right that these questions are out of scope for the PR, but they were inspired by it. (`conda install pyxrootd` would change quite a lot!) Maybe we should move that thread to the Scikit-HEP Google Group?

eduardo-rodrigues:
Also adding @henryiii as we talked about ROOT and conda very very recently, and he has been thinking about these matters. Again, this discussion is well worth having in the scikit-hep-forum list :-).

jpivarski:
Actually, let's move the discussion, with its broader implications, to an e-mail thread. On the admins list? Or the general list?

chrisburr:
Yeah I agree that this discussion should be continued on one of the lists.

@jpivarski This PR now uses conda to get Python for all the builds and includes ROOT where possible. I think the remaining failure is an actual bug.

eduardo-rodrigues:
I would go for the forum list, not the admins. There is nothing private in the discussion and the topic is of interest to many. Thanks.

henryiii:
Travis now recommends Xenial for general use, and that is the only platform they support Python 3.7 on. https://blog.travis-ci.com/2018-11-08-xenial-release. Even if we are just using conda, I'd probably go to Xenial.

Also, let's chat about the root-conda package. I've been working on something too and have been in contact with the ROOT team. I'm around at CERN if you'd like to chat in person. Let's use the scikit-hep-forum list.

jpivarski:
Be aware that you're going to have to make a fairly substantial merge. To accommodate [optional acceleration of awkward-array](https://github.com/scikit-hep/awkward-array/pull/70), a lot of [references to awkward and numpy](https://github.com/scikit-hep/uproot/pull/220) had to be moved.

It should be easy to merge setup.py and test_write.py, but .travis.yml will be more difficult. That's where you've done most of your work, and I also had to edit this because I've changed all awkward-array → uproot-methods → uproot test matrices from Pythons 3.4‒3.6 to 3.5‒3.7 (can't test both from the same Ubuntu distribution). Beyond the fact that it's a good idea to test 3.7 nowadays, Numba doesn't seem to support 3.4. (Also, the Python 3.7, Numpy 1.13.1 combination had to be explicitly excluded; it tries and fails to compile Numpy. And don't miss the `pip install --upgrade pyopenssl`, needed because deployment was broken without it in Xenial.)

This is just a heads-up. (I didn't want to make your PR more difficult, sorry!)

chrisburr:
Thanks for the heads up and it's okay, I knew this PR had rotted.

When you have time, could you take a look at the test I added [that fails with Python 2.7](https://travis-ci.org/scikit-hep/uproot/jobs/486519507#L695)? I'm not sure if it's a bug in uproot or a bug in the test.

jpivarski:
I didn't mean to accidentally close this PR, sorry.

chrisburr:
@jpivarski I think this is good to go, though Python 3.4 has been added to the test matrix. Shall I remove it again or limit it to just one version of numpy?

jpivarski:
The only reason I left it out was because of incompatibility with Numba in awkward-arrray, but that doesn't apply here because uproot doesn't use Numba. We might as well test it!

I've been going over the changes. It looks like you've merged up to master and the tests pass. As I understand it, awkward-array and uproot-methods (and a smattering of other dependencies) come from pip and ROOT comes from conda. Bringing in ROOT allows `test_strings` to verify that ROOT can read the string written by uproot. @reikdas and I can add a few more tests there!

I miss the tests having the Python version next to them— my screen isn't quite wide enough to see the Python and Numpy version in the environment variables section— but that's a quibble. The tests take longer, but still just a few minutes on average.

You think it's ready. It looks good to me, too. I'll press "merge" unless you stop me in the next couple of minutes.

chrisburr:
> I miss the tests having the Python version next to them

Yeah, I agree. I'll shorten the variable names.

chrisburr:
> Maybe you could instead label them with TRAVIS_JOB_NAME?

[Apparently not...](https://travis-ci.org/chrisburr/uproot/builds/486606941)

> As I understand it, awkward-array and uproot-methods (and a smattering of other dependencies) come from pip and ROOT comes from conda.

Yup, you could also use conda for everything but I preferred minimising how invasive this PR is.

jpivarski:
> > As I understand it, awkward-array and uproot-methods (and a smattering of other dependencies) come from pip and ROOT comes from conda.
> 
> Yup, you could also use conda for everything but I preferred minimising how invasive this PR is.

I prefer using pip for awkward and uproot-methods dependencies, too, because I can control when and how those two go up onto PyPI. I was just making sure I understand what's going into this PR.

If the tests pass, are you ready for me to merge? It looks good to me.

chrisburr:
Yup go ahead 👍

PR #184, opened 2018-11-19
Update ROOT version while writing

reikdas:
We should probably merge this in(with the then current version of ROOT), once we have the tests for writing.

### DIFF
--- a/uproot/write/TFile.py
+++ b/uproot/write/TFile.py
@@ -210,7 +210,7 @@ class TFileRecreate(TFileUpdate):
 
     def _writeheader(self):
         cursor = uproot.write.sink.cursor.Cursor(0)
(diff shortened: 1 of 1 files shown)


jpivarski:
Okay, I'll take this PR if this ROOT version is exactly equal to the one used to generate the streamers we write into ROOT files. There should be a one-to-one relationship between ROOT version and streamers. We shouldn't say that we're producing files consistent with the latest ROOT version if we're not, and we shouldn't expect to always be completely up-to-date with ROOT releases. (Only if we're updating something else anyway or there's a significant change we want to emulate.)

reikdas:
Ah, I am closing this then as the ROOT version changes should probably come in with the version update-requiring-change as a single pull request.

jpivarski:
Okay.

PR #189, opened 2018-12-07
Fix moved array types.

riga:
Hi Jim,

I ran into an error after a fresh install of uproot and awkward from latest pypi versions.
uproot complained about a missing `awkward.util.INDEXTYPE`.

I saw that you moved the types to the base AwkwardArray, and noticed that this change is not yet reflected here in uproot, so I did a quick replacement.

The changes in this PR use `awkward.JaggedArray` where jagged arrays are handled, and `awkward.array.base.AwkwardArray` otherwise.

### DIFF
--- a/uproot/interp/jagged.py
+++ b/uproot/interp/jagged.py
@@ -80,7 +80,7 @@ class asjagged(uproot.interp.interp.Interpretation):
         return awkward.type.ArrayType(awkward.util.numpy.inf, self.content.type)
 
     def empty(self):
-        return awkward.JaggedArray(awkward.util.numpy.empty(0, dtype=awkward.util.INDEXTYPE), awkward.util.numpy.empty(0, dtype=awkward.util.INDEXTYPE), self.content.empty())
+        return awkward.JaggedArray(awkward.util.numpy.empty(0, dtype=awkward.JaggedArray.INDEXTYPE), awkward.util.numpy.empty(0, dtype=awkward.JaggedArray.INDEXTYPE), self.content.empty())
 
     def compatible(self, other):
         return isinstance(other, asjagged) and self.content.compatible(other.content)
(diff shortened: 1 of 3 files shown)


jpivarski:
I've got a computer now.  :)

I just verified that this hits 100% of the `awkward.util.*TYPE` matches. I'll just add to this pull request to bump the version number.

jpivarski:
Oh. Well, I was just looking at it and thinking I'd change the `awkward.array.base.AwkwardArray.CHARTYPE` to `awkward.ObjectArray.CHARTYPE` because those instances (in objects.py) are making `ObjectArrays`.

It reminds me that uproot ought to start using the `StringArray` class, which was created to handle strings from Parquet and Apache Arrow. It's hard to wrangle these cats.

I'll check the version number and just commit the change I described in this comment. It's going into your fork, anyway.

jpivarski:
Yes, that's the right version number. Here's the `CHARTYPE` switch.

jpivarski:
Actually, this implies a strict dependency on the version of awkward. uproot now depends on `awkward >= 0.5.6`, so I'm setting the dependencies in setup.py and requirements.txt to reflect that.

I knew that version matching would become an issue by splitting the repositories, but there are advantages to that as well. The reason that all arrays are referenced as being at `awkward.JaggedArray` etc. rather than `awkward.array.jagged.JaggedArray` was to give me the freedom to organize things internally at the cost of necessitating that they're always visible at top-level. But I think users would want them at top-level, anyway, so that was the trade-off.

(I've already moved `StringArray` as part of an internal reorganization...)

sbinet:
just a drive-by comment but, you should probably bump to `0.6.0` to follow correct SemVer semantics.

jpivarski:
That was my plan, actually. That's why I didn't update uproot; I was going to do that at the next minor number, awkward 0.6.0. But this PR came in, and I decided to do it now because of the interest, then thought through the consequences, that it would require pinning to a version number like 0.5.6.

PR #196, opened 2018-12-14
More histogram intro

HDembinski:
Minimal extension of the README, which shows how to get the most important data of a histogram. Perhaps some link to uproot-methods should be added as well, to point to further information.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -881,15 +881,19 @@ We have to add functionality by writing the equivalent Python. The uproot `TTree
     [3, inf]     0    |                                                            |
                       +------------------------------------------------------------+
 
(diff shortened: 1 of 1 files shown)


jpivarski:
Looks good. Is the `h.variances` in that example supposed to be empty because there's no `_fSumw2`? Should it perhaps be `None` or `numpy.sqrt(h.values)` instead?

jpivarski:
If the `h.variances` interface changes, the README will have to change, but I'm leaving that for the future.

jpivarski:
Ah wait— I needed to read that in context. I'll edit on master (it's a README).

jpivarski:
I misspoke: `variances` would be equal to `values` without a `_fSumw2`, not the square root. (I'm thinking "errors.") I think we're in agreement, then.

HDembinski:
Yes. :)

PR #203, opened 2019-01-07
Revert "Use Binder build API and remove trigger_binder.py and selenium dependency"

jpivarski:
Reverts scikit-hep/uproot#202

PR #202, opened 2019-01-07
Use Binder build API and remove trigger_binder.py and selenium dependency

matthewfeickert:
This PR is the effectively the same as [`pyhf` PR 383](https://github.com/diana-hep/pyhf/pull/383).

To summarize it here as well:

```
* Use the Binder build API endpoint with Python's built in webbrowser to trigger builds
   - https://gitter.im/jupyterhub/binder?at=5c2f87038dafa715c73ff54f
   - https://github.com/jupyterhub/binderhub/blob/9ca8fa68bb8b69c6a2736f2275279583073f314f/examples/binder-api.py#L28
* Remove selenium and chromium-chromedriver dependency installation from Travis CI docs stage
* Remove binder/trigger_binder.py as being able to access the API makes it unnecessary
```

This effectively reverts parts of PR #149 and PR #153

PR #207, opened 2019-01-14
Use Binder build API and use curl to trigger build

matthewfeickert:
As a follow up/fix to PR #202 and a reversion of PR #203 this reapplies the removal of `binder/trigger_binder.py` and its associated selenium and chromium-chromedriver dependencies. It additionally uses `curl` instead of `python -m webbrowser` to ping the Binder build web API to avoid timeout issues and to use the simplest tool possible for a simple job.

```
* Use the Binder build API endpoint with curl to trigger builds
   - https://gitter.im/jupyterhub/binder?at=5c2f87038dafa715c73ff54f
   - https://github.com/jupyterhub/binderhub/blob/9ca8fa68bb8b69c6a2736f2275279583073f314f/examples/binder-api.py#L28
* Remove selenium and chromium-chromedriver dependency installation from Travis CI docs stage
* Remove binder/trigger_binder.py as being able to access the API makes it unnecessary
* Rename stage to binder
```


matthewfeickert:
As an example of `curl` working with a new branch and not timing out [c.f. Travis CI build 1390.1](https://travis-ci.org/scikit-hep/uproot/jobs/479246330)

matthewfeickert:
> I tried to trigger a build to see if it would do the "binder" (used to be "docs") section, but it won't

You mean you hit the API with `curl` at your command line? If you saw something like "found built image" that's because I have been hitting the API for the last couple of days and so it has already built the image for the commit that was at the `HEAD` of `master`. Or do you mean something else?

jpivarski:
I mean that it created a job without the "binder" section, so it won't be a proper test. Now that I've merged the PR and it's testing the master branch, the new test has a "binder" section, so that'll test it in vivo. I'm keeping this tab open while that test is running, so that I can revert it if it fails.

matthewfeickert:
> I mean that it created a job without the "binder" section, so it won't be a proper test. 

Oh yeah. That's why I ran [the test in a test branch in my first comment of this PR which does](https://travis-ci.org/scikit-hep/uproot/jobs/479246330)

PR #211, opened 2019-01-16
Http authentication

### LINKED ISSUE #210: read password protected fileover http
I'm trying to read a file via http:
```python
file = uproot.open("http://data.magic.pic.es/Data/SuperStar/v1/CrabNebula/2014_01_01/20140101_05032025_S_CrabNebula-W0.40+035.root")
```
I get the following error:
```python
HTTPError: HTTP Error 401: Unauthorized
```
Which is basically what I'm expecting, because if I try to download it directly with my browser, I'm prompted
> http://data.magic.pic.es is requesting your username and password. The site says: “Authenticated proxy”

Is there a way to read a file via http on a server which is password protected?

noahhdf:
Resolves #210 

As suggested in #210 I implemented http authentication.
On the way I switched from urllib to [`requests`](http://docs.python-requests.org/en/master/).

Now i can do this:
```python
url = "http://data.magic.pic.es/Data/SuperStar/v1/CrabNebula/2014_01_01/20140101_05032025_S_CrabNebula-W0.40+035.root"
file = uproot.open(url, httpsource={"auth": ("username", "password")})
```
raising an `HTTPError` if the authentication is missing or wrong.

Opening non-protected files is the same as before.

The syntax for [authentication](http://docs.python-requests.org/en/master/user/authentication/#basic-authentication) is copied from requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,8 @@ install:
   - pip install "uproot-methods>=0.3.0"
   - python -c 'import uproot_methods; print(uproot_methods.__version__)'
   - pip install pytest pytest-runner
+  - pip install lz4 requests
+  - if [[ $TRAVIS_PYTHON_VERSION = "2.7" ]] ; then pip install backports.lzma ; fi
   - if [[ $TRAVIS_PYTHON_VERSION != pypy* ]] ; then pip install pandas ; fi
 
 addons:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -48,5 +48,5 @@ build_script:
   - "pip install awkward>=0.7.0"
   - "pip install uproot-methods>=0.3.0"
   - "pip install -i https://pypi.anaconda.org/carlkl/simple backports.lzma"
-  - "pip install pytest pytest-runner pandas"
+  - "pip install pytest pytest-runner pandas requests"
   - "python setup.py pytest"
--- a/binder/requirements.txt
+++ b/binder/requirements.txt
@@ -6,3 +6,4 @@ pkgconfig
 lz4
 backports.lzma;python_version<"3.3"
 pandas
+requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ awkward>=0.7.0
 uproot-methods>=0.3.0
 cachetools
 backports.lzma;python_version<"3.3"
+requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@ awkward>=0.7.0
 uproot-methods>=0.3.0
 cachetools
 backports.lzma;python_version<"3.3"
-requests
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ setup(name = "uproot",
       download_url = "https://github.com/scikit-hep/uproot/releases",
       license = "BSD 3-clause",
       test_suite = "tests",
-      install_requires = ["numpy>=1.13.1", "awkward>=0.7.0", "uproot-methods>=0.3.0", "cachetools"],
+      install_requires = ["numpy>=1.13.1", "awkward>=0.7.0", "uproot-methods>=0.3.0", "cachetools", "requests"],
       setup_requires = ["pytest-runner"],
       tests_require = ["pytest", "pkgconfig", "lz4", 'backports.lzma;python_version<"3.3"'],
       classifiers = [
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ setup(name = "uproot",
       download_url = "https://github.com/scikit-hep/uproot/releases",
       license = "BSD 3-clause",
       test_suite = "tests",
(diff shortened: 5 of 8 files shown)


matthewfeickert:
@noahhdf This is cool. :+1:  Jim can review this, but what would be really nice is if you could also write some tests for this (both a failing and a non-failing case) since this is adding in new functionality.

noahhdf:
I‘ve written some local tests on my machine, I can add them to this PR. 
Writing a test for http access is easy with the data files provided in the Readme.

Password protected access is harder, because one would have to provide an access restricted server where everyone can have the password etc

noahhdf:
So I've pushed 5 tests for combinations of authentication

1. not needed, not given (ok)
2. not needed, but given (ok)
3. needed, but not given (not ok)
4. needed, and given correct authentication (ok)
5. needed, but given wrong authentication (not ok)

The problem so far is, that tests 3 and 5 will fail because I do not have the opportunity to set up a protected server

jpivarski:
Thanks for this!

I intend to add to this PR to make the requests library an optional dependency in the same sense as XRootD or LZ4— by burying the import in the HTTPSource constructor and printing a "how to install" pip/conda pair in the exception message. Not everybody uses HTTP, so requests shouldn't be a strict dependency for uproot, though admittedly, most users probably already have it.

As for tests, I don't want any of the default tests to depend on the network. Depending on too many things outside of itself makes the testing suite fragile. For the same reason, tests can't depend on random numbers or concurrency, even though concurrency would be a great thing to test.

Perhaps pytest has a way to label some tests "extra" and let the user opt in to them? If so, we should use that feature, and the tests you're discussing above would be the first under the "network" category of extra tests.

eduardo-rodrigues:
Yes it does, @jpivarski. I have used

```
# Obviously, skip all tests if the ROOT library is not available!
ROOT = pytest.importorskip('ROOT')
```

when I wanted to skip some tests if ROOT is not available. I'm sure there are extra possibilities but this is probably what you need.

matthewfeickert:
> Perhaps pytest has a way to label some tests "extra" and let the user opt in to them?

@jpivarski @eduardo-rodrigues I think (haven't used these myself a lot yet in my pytest projects) that what is wanted here is [`pytest.mark.skipif`](https://docs.pytest.org/en/latest/skipping.html).

eduardo-rodrigues:
Thanks @matthewfeickert, Indeed that and the skip (no "if") fixture can be really handy for the 2 tests that cannot run for the reasons mentioned above. With this fixture you can keep the  test in, not the least for "documentation".

matthewfeickert:
> Password protected access is harder, because one would have to provide an access restricted server where everyone can have the password etc

@noahhdf Maybe you could look into using [`monkeypatch`](https://docs.pytest.org/en/latest/monkeypatch.html) or [`mocker`](https://pypi.org/project/pytest-mock/)?

jpivarski:
`pytest.importorskip` is good for testing optional dependencies, like XRootD or LZ4. We can then ensure that these tests run on Travis by making explicitly installing those libraries in the Travis setup.

This is a little different: users probably already have the "requests" library installed, but that doesn't mean we want the tests (without the user opting in) to depend on the network. What we want is more like pytest's "markers": https://docs.pytest.org/en/latest/mark.html#mark

The pytest examples include markers for things like "slow tests" that you wouldn't want to run every time, but you can opt into them on the command line with `pytest -m slow`. That, I believe, would be the right balance. We'd want a `pytest -m network`.

Meanwhile, I'll make the edits I promised to this PR.

matthewfeickert:
> The pytest examples include markers for things like "slow tests" that you wouldn't want to run every time

As a compliment to this if there were tests that did mocking then the machinery of the code would still get tested in the normal running of the tests. As long as the payload doesn't matter then http requests can be mocked out very nicely.

jpivarski:
Yeah, if it's done with mocking, not a real HTTP connection, then there would be no reason to exclude it from the default tests. Whereas I'm dependency-averse for the main library, tests can have as many dependencies as they want, as long as they come from pip.

jpivarski:
I submitted a review that "requested changes" because I was requesting that the failing tests be made to pass. But then that "requested changes" blocker is redundant with the failing tests blocker, so I "approved" it for simplicity.

jpivarski:
@MaxNoe That's a good idea, though it might be exactly how one of these monkeypatch/mocker libraries work. Implementing it ourselves could get tricky when the server has to run in a different thread from the test, because they would need to be synchronized. Without sufficient care, it might work 99% of the time, but a randomly failing test would be an issue.

noahhdf:
I implemented mocking the `requests.get` function.

Now the tests that should raise an `HTTPError` do so by just bypassing the connection and getting the `status_code = 401` (for not authenticated).

The tests that should run open a local file isntead of an http connection and go on from there.

jpivarski:
@noahhdf This looks good to me! Are you ready for me to merge it?

jpivarski:
I just started the process to deploy this as uproot 3.3.3. Thanks!

Issue #217, opened 2019-01-23
Option in iterate method to access the current file used by uproot.iterate at each step


HealthyPear:
I open first an issue (actually more of a feature request) instead of directly create a new pull request, so we can discuss first.

I created a virtual environment with _pipenv_ to play a bit with the source code - this is my proposal:

1. make __iterate_ function return by default also the file path from line 162 `for path in paths:`, modifying line 183 as follows,
`yield tree, newbranches, globalentrystart ---> yield tree, newbranches, globalentrystart, path`
2. add a new _reportfile_ argument defaulting to _False_ to the definition of _iterate_,
3. retrieve _path_ from __iterate_, modifying line 125 as follows,
    `for tree, newbranches, globalentrystart in _iterate(.....): ---> for tree, newbranches, globalentrystart, path in _iterate(.....):`
4. add a check to the existing one on `reportentries` in order to take into account all 4 possibilities (both True, only one True and both False), modifying lines from 130 to 133 as follows,
```
    if reportentries:
        yield globalentrystart + start, globalentrystart + stop, arrays
    else:
        yield arrays
```
--->
```
    if reportentries:
        if reportfile:
             yield globalentrystart + start, globalentrystart + stop, arrays, path
        else:
             yield globalentrystart + start, globalentrystart + stop, arrays
    else:
       if reportfile:
             yield arrays, path
       else:
             yield arrays
```
so path will always be the last argument.

I tested such proposed solution successfully on my setup using Python2.

jpivarski:
Wow, you did nearly all the work on this— I had mentally filed it away as future work to do.

I'm going to ask, though, that you turn it into a pull request:

1. [Fork](https://help.github.com/articles/fork-a-repo/) the uproot repo in GitHub.
2. Make the changes in [a new branch](https://help.github.com/articles/creating-and-deleting-branches-within-your-repository/) of your repository, using your virtualenv to check that they're properly applied.
3. [Pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) your fork to my repo, which I can then use to integrate them into a new version.

A pull request is an easier and safer way to communicate code changes (and you'll probably use it on other projects, if you're not already). If you do the pull request from a non-master branch of your repo, I can send my edits back, so we can converge on it. The pull request will also trigger the full testing matrix, so it will be tested by Travis for all versions of Python and all versions of Numpy. It's a good habit to get into.

Then we can get this done by the end of the day.

HealthyPear:
[...]

PR #218, opened 2019-01-23
Applied changes in issue #217 of original repository scikit-hep/uproot.

HealthyPear:


jpivarski:
Congrats on your first public pull request! I've made a few edits:

   * don't overshadow the `path` variable (introduced a new name, `thispath`)
   * added `reportfile` (get file objects) to `reportpath` (get path strings) so that the user doesn't have to reopen files (possibly many times during the processing of a single file), but `reportpath` is still available for that convenience
   * generalized `yield` processing, so it's clear where to put more of these in the future, should they be added
   * added help support

The order is also different from how you set it up, so you'll need to adjust your (less than a day old) scripts. These "reports" should either all go before the arrays or all go after the arrays, not a mix, and `reportentries` has already set the convention to _before_.

The tests are running. Let me know at any time if you approve these changes; if you do, I'll merge it in.

jpivarski:
Silence implies consent: if I don't hear from you, I'll just merge this PR.

HealthyPear:
Hello,

actually, silence can mean also that I couldn't answer in time :)
I apologize for the delay, I just had the opportunity to modify the code by adding your edits...

I confirm that the current behavior is compatible with the last commit and that I am also able to see the file object printed when enabling "reportfile".

Thank you,
Best regards

jpivarski:
Sorry— I was a little too glib.  :)  Anyway, it should be posted to PyPI for "pip install" as version 3.3.5 shortly. Thanks for this new feature!

PR #220, opened 2019-01-29
Depend on awkward 0.8.0, uproot-methods 0.4.0

jpivarski:
Bringing everything up to date.


jpivarski:
I just tested the two tutorials by hand with the latest awkward and uproot-methods. I can't find any errors!

PR #222, opened 2019-01-31
eliminate Python 3.7 warning and cleanup

jpivarski:


jpivarski:
This is also a full test of the new deployment framework, but on a low-priority bug fix.

PR #223, opened 2019-02-06
Speedup TTree.pandas.df(flatten=True) by about factor 8

guitargeek:
Hi Jim!

I am working a little bit with the data analysis for the beam tests of the CMS endcap calorimeter upgrade, and we have ROOT files of the following pattern:
* Some branches with basic data types (`int`, `float`)
* Some branches with vectors, all the same length per event

I think the `TTree.pandas.df(flatten=True)` functionality is intended for this use case, but this is about 8 times slower than the equivalent functionality in root_pandas, which everyone therefore wants to use and has to go through the pain of installing ROOT. I realized the bottleneck of uproot is here in the merge:

https://github.com/scikit-hep/uproot/blob/master/uproot/_connect/to_pandas.py#L194

Pandas is very slow at merging DataFrames with MultiIndex and the normal Index together, which is happening there: for the basic type branches uproot creates simple DataFrames, for the vector branches DFs with MultiIndices.

It could be significantly sped up by turning the basic type branches into the correct structure by adding them to a JaggedArray of zeros (the broadcasting is just too useful). Additionally, if the broadcasting (or flattening for the awkward arrays) is done in a separate loop over the branches, one can save some code duplication when filling the data frame.

Therefore, this PR should still cover all usecases for `TTree.pandas.df(flatten=True)` but just be much faster. Actually a few ten % faster than root_numpy now. I tested it with an ntuple from the testbeam, with I uploaded here:
https://rembserj.web.cern.ch/rembserj/dumps/ntuple_426.root

What do you think about this? You think it goes in the right direction?

These are the speed benchmarks I did:

```Python
from root_pandas import read_root
import uproot

branches = [u'event', 
       u'rechit_layer','rechit_module', u'rechit_chip', u'rechit_channel', u'rechit_x', u'rechit_y', u'rechit_z',
       u'rechit_energy', u'rechit_amplitudeHigh', u'rechit_amplitudeLow', u'rechit_Tot', 'rechit_toaRise',
       'rechit_TS3High','rechit_TS3Low',  'rechit_TS2High','rechit_TS2Low']

rh_branches = [branch for branch in branches if "rechit" in branch]

%%time
df_root_pandas = read_root("ntuple_426.root",key='rechitntupler/hits',
                           columns=branches, flatten = rh_branches)

%%time
df_uproot = uproot.open("ntuple_426.root")["rechitntupler/hits"].pandas.df(
    branches=branches, flatten=True)
```

Shout-out to @artlbv who originally recognized the slowness and made the comparison with root_pandas!

### DIFF
--- a/uproot/_connect/to_pandas.py
+++ b/uproot/_connect/to_pandas.py
@@ -105,12 +105,11 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
         return outputtype(columns=columns, data=data)
 
     else:
-        scalars = []
 
         length = None
         starts, stops = None, None
 
-        unpacked_tuple = []
+        presents = []
 
         for future_tuple in futures:
             name, interpretation, future = future_tuple
@@ -133,24 +132,24 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
             else:
                 array = future()
 
-            unpacked_tuple.append((name, interpretation, array))
+            presents.append((name, interpretation, array))
+
+        names, interpretations, arrays = zip(*presents)
+        arrays = list(arrays)
+        lengths = [len(a) for a in arrays]
 
         index = pandas.Index(numpy.arange(length))
-        out = outputtype(index=index)
+        df = outputtype(index=index)
 
-        def align_by_broadcast(array):
-            # Invoke jagged broadcasting to align arrays
-            content = awkward.numpy.zeros(stops.max(), dtype=array.dtype)
-            return (awkward.JaggedArray(starts, stops, content) + array).flatten()
-
-
-        names, interpretations, arrays = zip(*unpacked_tuple)
+        for i in range(len(arrays)):
+            if len(arrays[i]) < max(lengths):
+                # Invoke jagged broadcasting to align arrays
+                content = awkward.numpy.zeros(stops.max(), dtype=arrays[i].dtype)
+                arrays[i] = (awkward.JaggedArray(starts, stops, content) + arrays[i]).flatten()
 
         for name, interpretation, array in zip(names, interpretations, arrays):
 
             if isinstance(interpretation, asjagged):
-
-                df = outputtype(index=index)
 
                 interpretation = interpretation.content
                 if isinstance(interpretation, asobj) and isinstance(interpretation.content, astable):
@@ -181,41 +180,31 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
 
                 else:
                     fn = flatname(name, None, ())
-                    df[fn] = list(array)
+                    df[fn] = array
 
             elif isinstance(interpretation, asdtype):
-                df = outputtype(index=index)
                 if interpretation.todims == ():
                     if interpretation.todtype.names is None:
                         fn = flatname(name, None, ())
-                        df[fn] = align_by_broadcast(array)
-                        scalars.append(fn)
+                        df[fn] = array
                     else:
                         for nn in interpretation.todtype.names:
                             if not nn.startswith(" "):
                                 fn = flatname(name, nn, ())
-                                df[fn] = align_by_broadcast(array[nn])
-                                scalars.append(fn)
+                                df[fn] = array[nn]
                 else:
                     for tup in itertools.product(*[range(x) for x in interpretation.todims]):
                         if interpretation.todtype.names is None:
                             fn = flatname(name, None, tup)
-                            df[fn] = align_by_broadcast(array[(slice(None),) + tup])
-                            scalars.append(fn)
+                            df[fn] = array[(slice(None),) + tup]
                         else:
                             for nn in interpretation.todtype.names:
                                 if not nn.startswith(" "):
                                     fn = flatname(name, nn, tup)
-                                    df[fn] = align_by_broadcast(array[nn][(slice(None),) + tup])
-                                    scalars.append(fn)
+                                    df[fn] = array[nn][(slice(None),) + tup]
 
             else:
                 fn = flatname(name, None, ())
-                df = outputtype(index=index, columns=[fn], data={fn: list(array)})
-                scalars.append(fn)
+                df[df] = array
 
-            out = pandas.merge(out, df, how="outer", left_index=True, right_index=True)
-
-        for n in scalars:
-            out[n].fillna(method="ffill", inplace=True)
-        return out
+        return df
--- a/uproot/_connect/to_pandas.py
+++ b/uproot/_connect/to_pandas.py
@@ -140,7 +140,6 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname, aw
                         raise ValueError("cannot use flatten=True on branches with different jagged structure; explicitly select compatible branches (and pandas.merge if you want to combine different jagged structure)")
 
                 array = array.content
-                length = len(array)
                 needbroadcasts.append(False)
 
             else:
--- a/uproot/_connect/to_pandas.py
+++ b/uproot/_connect/to_pandas.py
@@ -29,6 +29,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import itertools
+import functools
+import operator
 
 import numpy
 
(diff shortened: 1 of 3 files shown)


uproot/_connect/to_pandas.py
```diff
@@ -104,95 +105,82 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
...
                                     df[fn] = array[nn][(slice(None),) + tup]
-                                    scalars.append(fn)
 
             else:
                 fn = flatname(name, None, ())
-                df = outputtype(index=index, columns=[fn], data={fn: list(array)})
-                scalars.append(fn)
-
-            out = pandas.merge(out, df, how="outer", left_index=True, right_index=True)
+                df[df] = array
```
guitargeek: For the scalars I think the procedure with the `fillna` is not needed, since this operation broadcasts them over the whole column.

uproot/_connect/to_pandas.py
```diff
@@ -104,95 +105,82 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
...
+                array = future()
+
+            names.append(name)
+            interpretations.append(interpretation)
+            arrays.append(array)
+
+        if length is None:
+            length = len(arrays[0])
+
+        index = pandas.Index(numpy.arange(length))
```
guitargeek: For now, this is not a MultiIndex anymore (root_pandas also has not MultiIndex in this case), but it would be just as quick if it's brought back if you prefer it.

jpivarski:
**That's fantastic!** I didn't realize there was an issue, but I'm glad you found it.

You're right that merging was to deal with some branches being jagged and others not being jagged. Doing it with JaggedArray's broadcasting, rather than Pandas's merging, puts it more in our control anyway. A very recent PR, scikit-hep/awkward-array#77, speeds up a common case in `__array_ufunc__` (involved in broadcasting) by a factor of 2.

An optimization should keep the interface the same, so let's keep the MultiIndex in for now. (Who knows if someone relies on it?) If you don't like the MultiIndex, you can just

```python
df = tree.pandas.df(flatten=True)
df.index = pandas.RangeIndex(len(df))
```

(Also, `pandas.RangeIndex(len(df))` is smaller and faster than `pandas.Index(numpy.arange(len(df)))`.)

It looks like one of the unit tests (specifically testing `TTree.pandas.df`) is failing; I'll let you look into that. I didn't immediately see what was wrong.

Eyeballing the code, it looks good to me. Once the tests pass, I'll take it. (Oh, and please bump the version number, so that it'll be ready for a new release. Thanks!)

uproot/_connect/to_pandas.py
jpivarski: Oh, that's right. `fillna` was in response to a user request, but it makes the final table look more like a broadcasted frame, anyway. So that's great: another expensive operation doesn't need to be there.

uproot/_connect/to_pandas.py
jpivarski: Since the purpose of this PR is optimization, let's keep the interface unchanged. (See comment below.)

guitargeek:
Hi! I just understood the issue with the unit test, it's because some branches contain 2d numpy arrays which can't get broadcasted with the JaggedArray. I don't know if there is a solution to make it work somehow (you have an idea?), or if we have to fall back on the merging in this case.

I'll look into this tomorrow, and also put the MultiIndex back.

guitargeek:
Oh and somehow the problem with the columns randomly turning into floats is gone. I had it too before for our event index column. I suppose that's the advantage of as you say having it more under control instead of letting pandas do it!

uproot/_connect/to_pandas.py
```diff
@@ -104,95 +105,82 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
...
-                                        df[fn] = array[nn][(slice(None),) + tup].flatten()
+                array = future()
 
-                else:
-                    fn = flatname(name, None, ())
-                    df[fn] = list(array.flatten())
+                if starts is None:
+                    starts = array.starts
+                    stops = array.stops
+                array = array.flatten()
```
jpivarski: I guess my original code did it, but if we're taking `starts` and `stops` as-is, without changing them to `counts` and such, then we're already assuming that `array.content == array.flatten()`. This was naive in uproot-methods, where the arrays could have come from anywhere, but it's probably justified coming directly out of a ROOT file, where `array.content == array.flatten()` may be guaranteed. Since we're making the `starts`, `stops` assumption, we should probably be consistent and use `array.content` as well, just to make that assumption more explicit.

uproot/_connect/to_pandas.py
```diff
@@ -104,95 +105,82 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
...
+        index = pandas.Index(numpy.arange(length))
+        df = outputtype(index=index)
+
+        for name, interpretation, array in zip(names, interpretations, arrays):
+
+            if isinstance(interpretation, asdtype):
+
+                if len(array) < length:
+                    # Invoke jagged broadcasting to align arrays
+                    content = awkward.numpy.zeros(stops.max(), dtype=array.dtype)
```
jpivarski: Again with the `array.content == array.flatten()` assumption, `stops.max() == stops[-1]` and the latter is faster to compute.

uproot/_connect/to_pandas.py
jpivarski: I changed it back into a MultiIndex.

uproot/_connect/to_pandas.py
```diff
@@ -104,95 +105,82 @@ def futures2df(futures, outputtype, entrystart, entrystop, flatten, flatname):
...
+            length = len(arrays[0])
+
+        index = pandas.Index(numpy.arange(length))
+        df = outputtype(index=index)
+
+        for name, interpretation, array in zip(names, interpretations, arrays):
+
+            if isinstance(interpretation, asdtype):
+
+                if len(array) < length:
```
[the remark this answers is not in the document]
jpivarski: I removed the `if len(array) < length` (the code that was inside the block always runs now), but please tell me what you were thinking in case I'm misinterpreting you.

uproot/_connect/to_pandas.py
jpivarski: I turned `.flatten()` → `.content` and `stops.max()` → `stops[-1]`, strongly making the assumption that `starts` and `stops` are an offsets pair, since they just came from the ROOT file with no user intervention.

uproot/_connect/to_pandas.py
jpivarski: Resolved as described above.

jpivarski:
One capability that the Pandas-merging code had but the new code has lost was to include branches with different jaggedness in the same DataFrame: for instance, both muons and electrons (which you'd get if you don't specify branches). Pandas-merging aligned them as best it could (filling in with `fillna`, which isn't especially meaningful), but JaggedArray-broadcasting can't deal with that at all.

I put in a ValueError explaining the situation to the user. Merging different jaggedness in the same DataFrame didn't make a lot of sense, analysis-wise, so I think it's actually an improvement that this no longer happens by default. It's possible to explicitly call `pandas.merge(df1, df2, left_index=True, right_index=True)` to get the old behavior, but now users have to do that themselves. More than just being aware where there slowdown comes from, they'd be more aware of this structure-mixing and can only get it if they really want it.

uproot/_connect/to_pandas.py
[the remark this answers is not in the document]
jpivarski: Eventually, I added a `needsbroadcast` boolean to specify exactly when to broadcast and when not to, based on the original `interpretation` of the branch. Sorry if I wrote the original `len(array) < length` and I'm blaming it on you—if I did, it was a mistake that happened to work in most cases.

guitargeek:
Hi! Thanks in particular for your explanation about the broadcasting. And abut your other changes I really can't complain, with the `needsbroadcastings` you made much more explicit what I hacked together with the length comparison, which I see now can fail in the case you described.

And I'm very happy about your decision to raise an error when there is different jaggedness! I was already prepared to argue when you'd have pointed it out, but actually you summarized the arguments already two comments before and came to the same conclusion.

I tested again for our usecase, and it's still as fast after your changes. I just added one more commit removing a `print` and the little statement I previously put to get a default `length` in case there would not have been any jagged in the branches. I realized this case is already handled by line 68 in the same file.

I can't of anything more to check and agree it's ready to merge.

Shout-out to @artlbv who originally recognized the slowness and made the comparison with root_pandas!

artlbv:
Thanks a lot @guitargeek and @jpivarski for fixing this! It will be greatly appreciated ;)

On this comment:
> Merging different jaggedness in the same DataFrame didn't make a lot of sense, analysis-wise, so I think it's actually an improvement that this no longer happens by default.

I totally agree with both of you that handling various-length collections in the same dataframe is not the best practise. And that they could be just merged after some pre-analysis. It was more confusing when the flattening happened on both at the same time (or just failed as in `root_pandas`).

PR #229, opened 2019-02-11
Read Double32_t values with truncated mantissa

plexoos:
This commit implements unpacking of the Double32_t format in which the exponent
and mantissa of the original double value (cast to single precision) are
streamed as 8-bit and 16-bit integers respectively.

Depending on the specification provided by the user the ROOT's Double32_t
values can be packed and streamed to a ROOT file in other two formats. In the
first mode the original double precision values are simply cast to a single
precision and streamed without further manipulation. In the second mode the
double values (again, cast to float and scaled to a given range) are saved as
32-bit integers. These two modes are already taken care of by the
`uproot.interp.auto.interpret()` function and the
`uproot.interp.numerical.asdouble32` interpretation.

For details on Double32_t streaming modes see
https://root.cern/doc/v616/classTBufferFile.html#a44c2adb6fb1194ec999b84aed259e5bc

### DIFF
--- a/uproot/interp/numerical.py
+++ b/uproot/interp/numerical.py
@@ -287,13 +287,17 @@ class asdouble32(_asnumeric):
         return quotient
 
     def fromroot(self, data, byteoffsets, local_entrystart, local_entrystop):
+        # Make sure the shape of the interpreted data conforms to the shape of the input data
+        def reshape(array):
+            product = int(self.awkward.numpy.prod(self.fromdims))
+            quotient, remainder = divmod(len(array), product)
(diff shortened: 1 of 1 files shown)


jpivarski:
Thanks! This was something I didn't know about; your PR makes the implementation mature. I'll accept this when all tests pass. (I've reviewed it; looks good!)

plexoos:
Great! The need for this is dictated by our experiment (STAR@BNL) where we save track parameters with limited precision

jpivarski:
@plexoos If you're happy with it, I'm happy with it.

jpivarski:
(Our messages crossed: I was asking if you're ready for me to press "merge.")

PR #230, opened 2019-02-19
Add connection period and duration limits to Binder build API curl

matthewfeickert:
The use of `connect-timeout` is to make sure that the curl doesn't try to repeatedly hit the URL if the connection is down or unresponsive for an unreasonably long time. The use of max-time is so that for longer builds the curl doesn't try to follow them for the duration. It just hits the API endpoint and then exits at most 60 seconds later. Additionally the trigger function accepts `--max-time` exit return code, `28`, as valid and then return success.

An [example of this working on a test branch can be found on Travis CI Build #1487](https://travis-ci.org/scikit-hep/uproot/jobs/495674517).

Also [update the launch Binder badge](https://discourse.jupyter.org/t/help-us-choose-an-updated-launch-binder-badge/100).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,4 +96,4 @@ jobs:
     install: skip
     script:
     # Use Binder build API to trigger repo2docker to build image
-    - curl https://mybinder.org/build/gh/scikit-hep/uproot/master
+    - bash binder/trigger_binder.sh https://mybinder.org/build/gh/scikit-hep/uproot/"${TRAVIS_BRANCH}"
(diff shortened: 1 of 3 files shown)


jpivarski:
Also let me know if you're done, if I should merge as soon as these tests pass.

matthewfeickert:
@jpivarski 

> This sounds like a good idea and the shell script looks good (I can't say that it's bug-free by looking at it).

I'll never claim that it is robust against all pathologies, but this shell script is the same as what I use in 2 other projects and they seem to be working as expected.

> I saw the discussion about a new Binder launch button, but I don't know which option this refers to. It's not the huge, wide one with the server name and everything, right? This is one of many badges.

This just changes the badge to be blue (as it is now the new default from Binder). So it is truly just superficial.

> Also let me know if you're done, if I should merge as soon as these tests pass.

This is GTG in my mind. So feel free to merge whenever.

jpivarski:
Thanks!

PR #235, opened 2019-02-27
Test auto interpretation for branches in Event.root

plexoos:
The idea is to add interpretations for branches where it is currently missing


jpivarski:
Cool! I am in favor!

plexoos:
I'd like to push a few more commits to this PR with the intention of adding a new branch in test/samples/Event.root. That branch would model a structure we have in our DST trees which could not have been properly interpreted

jpivarski:
I'm very happy that you're doing this; the interpretations part of uproot is the roughest edge, since there's a wide-open field of things to know about how ROOT serializes objects.

Please do add a file with the kinds of branches you need to interpret. Perhaps it should be a separate file from Event.root, since this is a standard from ROOT's own unit tests, and if like to keep it identical to the original. Also, it's a binary file in git (which I'll be soon replacing with references to scikit-hep-testfiles).

plexoos:
I believe the existing Event.root was generated using the same code from ROOT. I would like to extend their example with the following data member in Track: 3fecffe993
Please let me know if a separate file should be produced or we can replace the current Event.root. I am ok with either way you prefer.

jpivarski:
My original thinking was that the one new branch should be in a new file. Since the branch is part of a class, that could be difficult, so I take that back. Having a whole copy of Event.root with one additional branch would be okay if it weren't the largest by far (30 MB), which bogs down git.

So, let's replace it.

Soon, all of these files (and the macros that produced them) will move to https://github.com/scikit-hep/scikit-hep-testdata, but I'll do that after yours is fully merged. (I really hope that solves the slow git-clone problem!)

Either let me know when you're ready for me to press "merge" or use a "[WIP]" in the PR name. (We have a tool that prevents me from accidentally merging while "[WIP]" is still in the name, but since not everybody knows about this, I wait for some sort of "okay, I'm done!" comment before merging.)

plexoos:
Thank you for the information @jpivarski 
Please merge this branch if no further comments and when the tests pass.

I would like to propose a solution to the newly introduced test branch but want to do it in a separate PR

jpivarski:
I think the Python 3.6, 3.7 failures are not related to your work: I think Travis is missing packages. I'll try adding that to the .travis.yml of this PR.

jpivarski:
Whew! I'll find out if this PyOpenSSL thing is still an issue when I try to deploy it now.

Issue #238, opened 2019-02-28
KeryError from numentries when branch is missing


belforte:
when ise on a root file w/o the requested branch, a KeyError is rised.
Here's an example with stack trace.

```
In [21]: nentries = uproot.numentries(dsLFNs, "Events", total=False )
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-21-e1e5fb76980a> in <module>()
----> 1 nentries = uproot.numentries(dsLFNs, "Events", total=False )

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/tree.pyc in numentries(path, treepath, total, localsource, xrootdsource, httpsource, executor, blocking)
   1592     else:
   1593         paths = [y for x in path for y in _filename_explode(x)]
-> 1594     return _numentries(paths, treepath, total, localsource, xrootdsource, httpsource, executor, blocking, [None] * len(paths))
   1595 
   1596 def _numentries(paths, treepath, total, localsource, xrootdsource, httpsource, executor, blocking, uuids):

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/tree.pyc in _numentries(paths, treepath, total, localsource, xrootdsource, httpsource, executor, blocking, uuids)
   1629     if executor is None:
   1630         for i in range(len(paths)):
-> 1631             _delayedraise(fill(i))
   1632         excinfos = ()
   1633     else:

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/tree.pyc in _delayedraise(excinfo)
     94         cls, err, trc = excinfo
     95         if sys.version_info[0] <= 2:
---> 96             exec("raise cls, err, trc")
     97         else:
     98             raise err.with_traceback(trc)

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/tree.pyc in fill(i)
   1618                 source = file._context.source
   1619                 file._context.classes["TTree"] = _TTreeForNumEntries
-> 1620                 out[i] = file[treepath]._fEntries
   1621                 uuids[i] = file._context.uuid
   1622             except:

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/rootio.pyc in __getitem__(self, name)
    234 
    235     def __getitem__(self, name):
--> 236         return self.get(name)
    237 
    238     def __len__(self):

/uscms/home/belforte/.local/lib/python2.7/site-packages/uproot/rootio.pyc in get(self, name, cycle)
[...]

PR #239, opened 2019-02-28
Just take the first commit of #237 because the rest won't work until scikit-hep-testdata is fixed

jpivarski:
Also, I'll be putting Stefano's bug-fix (#238) and feature request (`__contains__` for `ROOTDirectory`) here, since that's what motivated the cherry-picking of #237.


jpivarski:
This failed because we also need the HTTP 504 insensitivity. I'll hand-weave that in, too.

PR #241, opened 2019-03-01
Deal with fixed size arrays as part of jagged one

plexoos:
This is a proposed fix for reading a branch introduced earlier in the Event.root test

### DIFF
--- a/tests/test_tree.py
+++ b/tests/test_tree.py
@@ -640,7 +640,6 @@ class Test(unittest.TestCase):
             b'fTracks.fVertex[3]',
             b'fTracks.fPointValue',
             b'fTracks.fTriggerBits.fAllBits',
-            b'fTracks.fTArray[3]',
             b'fTriggerBits',
             b'fTriggerBits.TObject'
         ]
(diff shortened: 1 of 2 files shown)


plexoos:
I agree we should have such test. I was actually thinking about a bit more comprehensive one. What about reading values into a dataframe and compare them directly via df.equals() to a reference? Ideally, we would write a tree with uproot and read it back, then compare. But since the writing functionality is not available we can save the dataframe to a hdf5 file and read it back from there. The hdf5 format preserves all types of the dataframe columns.

jpivarski:
It doesn't have to involve writing—the feature you've added is a reading feature, so it can just be

```python
assert tree.array("branch", entrystop=10).tolist() == [...]
```

where `...` is whatever you get from the first 10 events.

In fact, even if the writing feature existed, it wouldn't be a unit test to check both in the same test—it would be an integration test and we'd still need a unit test to narrow in on an error should one occur.

plexoos:
Feel free to modify as you wish

jpivarski:
Perfect! Do you want me to accept this now or were you planning more for this PR?

plexoos:
I don't have anything to add to this PR. Please merge. Thank you!

PR #244, opened 2019-03-06
Open Binder into JupyterLab

matthewfeickert:
Following the [JupyterLab + Binder example repo](https://github.com/binder-examples/jupyterlab) open up the notebooks in Binder in an instance of JupyterLab instead of Jupyter server.

Here's the modified link: [Binder](https://mybinder.org/v2/gh/scikit-hep/uproot/master?urlpath=lab/tree/binder%2Ftutorial.ipynb)

I think it looks quite nice.


jpivarski:
I'm done with this. I'll merge it in as soon as the tests pass.

matthewfeickert:
@jpivarski if you want to keep it that's fine, but the "Limit the output so that it's not huge." commit isn't really necessary, as you can just right click on those cells and select "Enable Scrolling for Outputs"

scrolling

jpivarski:
What I tried was explicitly turning on scrolling for that one output (using a menu option like that), then saving the .ipynb file, hoping that that metadata would be included in the file. It wasn't, or at least, the JupyterLab on Binder didn't present that output as scrollable without user intervention. I wouldn't want users to have to turn on scrolling themselves, and anyway, the GitHub view has no such option, so it made more sense (pedagogically) to just limit the for loop.

Thanks for the heads-up, though!

PR #254, opened 2019-03-13
Fixes #253

### LINKED ISSUE #253: dev - context is not None for compressed ROOT files
The line - https://github.com/scikit-hep/uproot/blob/master/dev/streamergen.py#L51 cannot pass None as context for compressed ROOT file (in the case allstreamers.c - https://github.com/scikit-hep/uproot/blob/master/dev/allstreamers.c generates a compressed ROOT file). 

This should not matter if we are compressing the streamers while creating a file, but it can't hurt to generalise the script.

reikdas:
Seems like it was a trivial tweak.

### DIFF
--- a/dev/streamergen.py
+++ b/dev/streamergen.py
@@ -48,7 +48,7 @@ for x in f._context.streamerinfos:
     if data[x._fName.decode("ascii")] != x._fClassVersion:
         print("Old {0} version = {1}. New {0} version = {2}".format(x._fName, data[x._fName.decode("ascii")], x._fClassVersion))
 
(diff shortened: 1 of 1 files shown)


reikdas:
> First: has this changed? Are we inserting compressed streamers into ROOT files now?
No. 
 
> If we're not inserting compressed streamers into ROOT files, then streamergen.py must not accept compressed streamers, because then the wrong thing would be inserted into the ROOT files.
I understand. I'll close the issue. 

> What motivated issue #253?
I was testing writing compressed streamers (but decided against it - it is not required) and thought it might be good to have the ability to extract compressed streamers if required in the future (but as you pointed out, it would be a bug if we did).

Issue #256, opened 2019-03-15
Index of tree.pandas.df doesnt use entrystart for flat arrays


benkrikler:
If `tree.pandas.df` is used with `entrystart > 0` and a jagged-array is one of the requested branches, then the resulting index (with the default `flatten=True` option) is a multindex with entry and subentry columns and entry values ranging  from `entrystart` to `entrystop`.  This is different to if only flat branches are requested, however, the index is just the row number going from `0` to `entrystop - entrystart`:

```python
In [0]: 
import uproot
tree = uproot.open("CMS_HEP_tutorial_ww.root")["events"]
```
```python
In [1]: tree.pandas.df(["NMuon", "NElectron"], entrystart=100, entrystop=200).head()
Out[1]: 
   NMuon  NElectron
0      1          0
1      1          0
2      1          0
3      1          0
4      1          0
```
```python
In [2]: tree.pandas.df(["NMuon", "Muon_Px"], entrystart=100, entrystop=200).head()
Out[2]: 
                NMuon    Muon_Px
entry subentry                  
100   0           1.0 -45.877728
101   0           1.0 -47.455692
102   0           1.0  54.362259
103   0           1.0 -30.484287
104   0           1.0   6.948876
```

I was wondering if we could be an optional argument that makes the index numbering the same in both cases?

I've had a look, and I think it would just require a couple of lines of extra code above [Line 109 of _connect/to_pandas.py](https://github.com/scikit-hep/uproot/blob/master/uproot/_connect/to_pandas.py#L109), using [pandas.RangeIndex.from_range](http://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.RangeIndex.from_range.html#pandas.RangeIndex.from_range) along the lines of:
```python
index = pandas.RangeIndex.from_range(range(startentry, stopentry))
```

If you think this would be reasonable to add, I'd be happy to put together a PR along these lines.

jpivarski:
Thank you—that would be great. This definitely looks like an oversight. (The Pandas output code was revamped several times; it took us a while to learn how to use it properly.)

benkrikler:
That was fast :zap: !  I'll try and open something.

jpivarski:
You fixed this, I merged it, so I'll close this issue.

PR #257, opened 2019-03-15
Let flat pandas dataframes have the same index format as jagged dfs

benkrikler:
Adds an argument to futures2df function to respect entrystart in output dataframe.  Although it's an optional argument for futures2df, I'm not actually passing it through from `pandas.df` or `arrays` at this point, so it's essentially always enabled (since that's the default parameter).   I can set this up, but I was unsure if we want to add this argument to `arrays` since it's only meaningful if the outputtype is a pandas dataframe.

Addresses issue #256.

(Apologies, I accidentally pushed this as a new branch to the uproot repo, rather than my own fork -- happy to close, delete branch, and re-open from my fork.)

### DIFF
--- a/tests/test_tree.py
+++ b/tests/test_tree.py
@@ -665,3 +665,11 @@ class Test(unittest.TestCase):
             tree = uproot.open("tests/samples/HZZ-objects.root")["events"]
             tree.pandas.df("muonp4")
             tree.pandas.df("muonp4", flatten=False)
+            df = tree.pandas.df("eventweight", entrystart=100, entrystop=200)
+            index = df.index.tolist()
+            assert min(index) == 100
+            assert max(index) == 199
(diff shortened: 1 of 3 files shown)


benkrikler:
Ok I'll change that.  Do you want me to extend the pandas unit test in test_tree.py as well?

jpivarski:
Yes, thanks for the test!

Do you want me to deploy a new version to PyPI with this fix? If so, just bump the revision number and I'll deploy it.

Also, let me know explicitly when you're done. I don't want to merge something you're still working on.

benkrikler:
Ok, I've bumped the patch version.  I think this is all done from my side, so feel free to merge when you're happy.

jpivarski:
Okay, when the tests pass, I'll deploy it. Thanks again!

PR #258, opened 2019-03-16
Enable ZLIB and LZMA compression while writing

### LINKED ISSUE #181: Enable compression while writing
We set fCompress = 0 while creating a TFile using uproot - This means that the file isn't compressed. We need to add support for writing compressed ROOT files.

reikdas:
Fixes #181 
I have only tested creating compressed ROOT files containing TObjStrings so far. Once that works perfectly I will start looking at files containing TH1s. 
ZLIB and LZMA compressed files can be read by uproot but not by ROOT. 
The current interface only allows the user to set the compression algorithm and compression level while creating the file and does not let the user change it after.


reikdas:
(Some notes for myself - )
As far as I understand, 
TObjStrings are not compressed 
THist is compressed 
Streamers are compressed

Current status - 
Waiting for a reply to this - https://github.com/python-lz4/python-lz4/issues/168 to figure out the lz4 checksum. 
Need to figure out how THist is compressed.
uproot is able to read (uproot created)zlib and lzma compressed files. (Not dealing with THist yet)
ROOT is able to read (uproot created)zlib and lzma compressed objects which have less 10000 bytes.

reikdas:
> @reikdas This might be helpful: https://github.com/python-lz4/python-lz4/blob/master/lz4/frame/__init__.py#L133-L136

I couldn't figure out how to get an 8 byte checksum from the python-lz4 package (issue - https://github.com/python-lz4/python-lz4/issues/168) and I decided to try using the xxhash package to calculate the checksum instead. (But it is failing due to checksum mismatch - I am still working on it)

Currently, ROOT is only able to read those ZLIB and LZMA compressed files (written by uproot) where the length of the object being compressed is less than 10000 bytes. @jpivarski Do you know why this might be?

jpivarski:
I'm glad you're testing large objects. The 9-byte header is 3-byte identity, 3-byte compressed size, and 3-byte uncompressed size. ([How it gets read.](https://github.com/root-project/root/blob/a9e7546bcfc5bea26776423cdb427d5f999559b9/core/zip/src/RZip.cxx#L160-L167)) Those 3-byte numbers have a maximum of `2**24`, an unusual number. If either the uncompressed or the compressed size exceeds this number, then you can only write a chunk that maximizes this size, followed by another chunk with the rest (or more chunks, if necessary). Each chunk is separately compressed.

The differences you're seeing among classes, whether TObjString is compressed or not, probably has more to do with the object sizes. If the compressed size is larger than the uncompressed size—which can easily happen if the object is small, due to the fixed size of the compression dictionary—it won't be stored in a compressed way. In fact, that's very important for the reading code (whether ROOT, uproot, groot, or root4j) to know if the buffer is compressed or not: our only indication is whether `fObjlen == fNbytes - fKeylen`.

A checksum from one library isn't going to work for another: they're all different kinds of hashes. It has to come from the LZ4 library. Are you having trouble calling the function at all or getting the right answer?

reikdas:
> A checksum from one library isn't going to work for another: they're all different kinds of hashes. It has to come from the LZ4 library. Are you having trouble calling the function at all or getting the right answer?

I don't think ROOT generates the checksum from the LZ4 library. They use the xxhash library - https://github.com/root-project/root/blob/master/core/lz4/src/ZipLZ4.cxx#L63 
According to ROOT, the checksum generated by uproot (xxhash) is not the right answer. 
This fails even in uproot. The uproot error is - `_block.LZ4BlockError: Decompression failed: corrupt input or insufficient space in destination buffer.`

> Those 3-byte numbers have a maximum of 2**24, an unusual number. 

10000 < 2**24 . So that is probably not the issue. 
Also, I am assuming the limit to be 10000 (with the extra digit causing the problem perhaps?) The closest object length I was able to create was ~11500 which failed when trying to be read by ROOT.

jpivarski:
Oh, it's supposed to be xxhash; okay.

I also checked the size and saw that it's less than `2**24`, but I didn't know if you were talking about the compressed or uncompressed size. I didn't see any hint of a sign bit in the code that I pointed you to, but I didn't mentally exclude the possibility.

For both issues, the thing to do is to see what ROOT does in both cases. Maybe the best testbed is to make TTree branches of char type, since you can more finely control both the size of the baskets (with [AutoFlush](https://root.cern.ch/doc/v614/classTTree.html#ad4c7c7d70caf5657104832bcfbd83a9f)) and its contents (because you fill exactly the bits you want to fill). If you've never made a TTree before, you know that you'll need to soon!

With a testbed like this, you can see where ROOT turns over from writing a single "page" of compressed data or multiple "pages" ("page" is a word I made up because I don't know what else to call it: a chunk of compressed data with a 9-byte header). You can do that for a compression type you understand, such as ZLIB. Also, you can separately test the LZ4 hash—I mean xxhash—by controlling the content of the data written to the TBasket.

Oh—maybe the issue with computing the hash is because the TKey is included... or not included... Because you remember how an object saved in a ROOT file has a TKey in the TDirectory and then also the same TKey right before the object? That might be included in the calculation of the hash or not, and maybe you've been testing the wrong case.

reikdas:
An update - I am waiting for answers to either https://github.com/python-lz4/python-lz4/issues/170 or https://stackoverflow.com/questions/56191696/what-is-lz4s-corresponding-python-api-function-for-the-c-apis-function-lz4-com

jpivarski:
If it's been this long, don't expect an answer. (Sometimes very old threads are answered, but you can't rely on it—it could happen any year now or never.) Is it the case that we can do gzip and lzma? If so, let's restrict the interface to only accept those two options and close this pull request, opening an issue for lz4 for later.

(A lot of users would be happy with just gzip—that already makes a difference.)

reikdas:
I moved lz4 compression into a different branch. 
For some reason, the Github files changed diff seems to show changes that have already been committed into master even after I fixed the merge conflicts.

jpivarski:
Once I've reviewed this, I plan to include it in uproot 3.6. Thanks!

jpivarski:
I think the differences you're talking about are because this branch is behind master (i.e. reverse-differences). I'll do a git merge.

PR #264, opened 2019-03-27
Fix iteration over DataFrames and provide more interfaces

### LINKED ISSUE #263: MultiIndex pandas dataframe from uproot.iterate
Hi! First of all thank you very much for this awesome package!

I have a question regarding MultiIndex pandas dataframes and uproot.iterate. When I open a ROOT file via uproot.open, I am able to select branches which contain JaggedArrays with the same dimensionality, and make them into a pandas dataframe with MultiIndex. For example:

```
mytree = uproot.open(myfile)["mytree"]
mytree.pandas.df(['muonPt', 'muonEta', 'muonPhi'])
```

Depending on the event, I can have (say) 0, 1, or 2 muons, so I can have accordingly 0, 1, or 2 subentries, and the resulting pandas dataframe reflects that.

But I would like to process several files with the same structure using uproot.iterate. I haven't found a way to make the pandas dataframe with MultiIndex by selecting the right branches from the iterate, e.g.:

```
for arrays in uproot.iterate(listoffiles, "mytree", ['muonPt', 'muonEta', 'muonPhi'], outputtype=pd.DataFrame, executor=executor):
         listofdataframes.append(arrays)
pd.concat(listofdataframes)
```

Without "flatten=True" in the iterate command above, the dataframes come out containing JaggedArrays, and I'm not sure how to turn those into a MultiIndex structure. If I do include "flatten=True", however, I get an error about incompatible dimensionalities:

> ValueError: Shape of passed values is (1, 1382), indices imply (1, 1466)

(I think this is because of the variable number of muons per entry). Is there a way to get the same behavior from uproot.iterate on many files, as I would from tree.pandas.df() on a single file?

Thank you!
Andre

jpivarski:
This PR fixes #263 and provides new methods and functions:

   * `tree.pandas.iterate` is like `tree.pandas.df` in that it sets some Pandas-friendly defaults, but on `tree.iterate`, rather than `tree.arrays`.
   * `uproot.pandas.iterate` sets those Pandas-friendly defaults on `uproot.iterate`.

Various bugs were fixed. For instance, `tree.iterate` really wasn't Pandas-ready: it had fallen considerably behind `tree.arrays`, but fortunately most of the Pandas-specific stuff is in a function call now, so no code duplication was needed to get `tree.iterate` up to speed and now it will inherit any future updates.

Also, `globalentrystart` setting in `uproot.iterate` was broken for Pandas DataFrames, both for `MultiIndex` and for `RangeIndex`. The original code seemed to be expecting `Int64Index` for some reason.

All in all, the combination of iterate + Pandas was just out of date with respect to the changes that have been more fully tested on arrays + Pandas.

### DIFF
--- a/uproot/__init__.py
+++ b/uproot/__init__.py
@@ -143,10 +143,6 @@ from uproot.write.TFile import TFileCreate as create
 from uproot.write.TFile import TFileRecreate as recreate
 from uproot.write.TFile import TFileUpdate as update
 
-# import uproot._connect.to_pandas
-# pandas = uproot._connect.to_pandas.Pandas
-# del uproot._connect.to_pandas
-
 from uproot.source.memmap import MemmapSource
 from uproot.source.file import FileSource
 from uproot.source.xrootd import XRootDSource
@@ -168,6 +164,23 @@ from uproot.interp.objects import STLMap
 from uproot.interp.objects import STLString
 asdebug = asjagged(asdtype("u1"))
 
+from uproot.source.memmap import MemmapSource
+from uproot.source.xrootd import XRootDSource
+from uproot.source.http import HTTPSource
+def iterate(path, treepath, branches=None, entrysteps=None, namedecode="utf-8", reportpath=False, reportfile=False, flatten=True, flatname=None, awkwardlib=None, cache=None, basketcache=None, keycache=None, executor=None, blocking=True, localsource=MemmapSource.defaults, xrootdsource=XRootDSource.defaults, httpsource=HTTPSource.defaults, **options):
+    import pandas
+    import uproot.tree
+    return uproot.tree.iterate(path, treepath, branches=branches, entrysteps=entrysteps, outputtype=pandas.DataFrame, namedecode=namedecode, reportpath=reportpath, reportfile=reportfile, reportentries=False, flatten=flatten, flatname=flatname, awkwardlib=awkwardlib, cache=cache, basketcache=basketcache, keycache=keycache, executor=executor, blocking=blocking, localsource=localsource, xrootdsource=xrootdsource, httpsource=httpsource, **options)
+
+from types import ModuleType
+pandas = ModuleType("uproot.pandas")
+pandas.iterate = iterate
+del ModuleType
+del iterate
+del MemmapSource
+del XRootDSource
+del HTTPSource
+
 # put help strings on everything (they're long, too disruptive to intersperse
 # in the code, and are built programmatically to avoid duplication; Python's
 # inline docstring method doesn't accept non-literals)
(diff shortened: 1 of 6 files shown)


afrankenthal:
Hi Jim, I just manually updated my uproot version to 3.4.16 following the master branch, and I'm running into this error now when using uproot.iterate:

> AttributeError                            Traceback (most recent call last)
> <ipython-input-23-c5215ead68b2> in <module>()
>       4 
>       5 listofarrays = []
> ----> 6 for arrays in uproot.iterate(listoffiles[0:2], "SREffi_dsa/cutsTree", ['recoPt','recoEta'],outputtype=pd.DataFrame, flatten=True, executor=executor):
>       7     listofarrays.append(arrays)
> 
> /uscms/homes/a/as2872/nobackup/conda/uprootgit/lib/python3.6/site-packages/uproot/tree.py in iterate(path, treepath, branches, entrysteps, outputtype, namedecode, reportpath, reportfile, reportentries, flatten, flatname, awkwardlib, cache, basketcache, keycache, executor, blocking, localsource, xrootdsource, httpsource, **options)
>     117             if getattr(outputtype, "__name__", None) == "DataFrame" and getattr(outputtype, "__module__", None) == "pandas.core.frame":
>     118                 if type(arrays.index).__name__ == "MultiIndex":
> --> 119                     index = arrays.index.levels[0].to_numpy()
>     120                     awkward.numpy.add(index, globalentrystart, out=index)
>     121                 elif type(arrays.index).__name__ == "RangeIndex":
> 
> AttributeError: 'Int64Index' object has no attribute 'to_numpy'
> 

I notice you mentioned something about Int64Index in the PR note, but I'm not sure how exactly this might relate?

Thanks!

jpivarski:
I chose the `index.to_numpy()` method over the `index.values` property, where `index` is a Pandas Index. It's possible that this is too new—you need to update your version of Pandas and I need to switch to the old interface for better support.

Does a new version of Pandas fix it? Even if it does, I'm going to want to support more than just the latest version...

jpivarski:
This method was introduced in Pandas 0.24.0. The new property is `array` (what I actually want, rather than `to_numpy()`, which might do a conversion) and the old property is `values`. I'll make a new version of uproot that tries `array` before `values`.

afrankenthal:
Cool, I'm trying to install pandas 0.24.2 (current one is 0.23.4). It's probably a good idea to have the latest pandas anyway.

afrankenthal:
Cool, that worked! I think you were faster than me and I ended up getting 3.4.17, so it works with pandas 0.24.2. I can also try with my other conda env that still has pandas 0.23.4 to see if 3.4.17 works as well.

jpivarski:
Running another version through is also a good test of Travis. It looks like the system is back up—mostly. Jobs start in decent time now, but "solving environment" for conda in Python 2.7 now takes longer than Travis is willing to wait. I'm thinking this issue is unrelated to the Travis outage; it just happened at the same time.

I don't like not being able to deploy whenever I want. (Grumble.)

afrankenthal:
I feel you! But apart from the conda delay it looks like this was a major Travis outage.

Anyway, I tested 3.4.17 with 0.23.4 and it seems to work as well! I also asked Alexx who manages a conda package in the LPC to update uproot when the latest release gets deployed, so others don't have to install their own env.

Thanks a lot for your help and quick problem-solving!

PR #265, opened 2019-03-28
Write tests

reikdas:


reikdas:
Tests will need to be restarted once https://github.com/scikit-hep/uproot-methods/pull/46 has been merged and released to PyPI.

jpivarski:
Thanks for this. Don't worry about the fact that Travis seems to be failing: they're having serious issues right now.

https://www.traviscistatus.com/

We just need to wait until they get through that. (They're referring to it as "slow starts," but they're also having problems on the VMs—conda installations fail to solve their dependency sets in the 10 minutes they give before killing the build. It is not our mistake.)

jpivarski:
Check the .travis.yml—you might need to wait for uproot-methods 0.4.6 to be in conda, but just PyPI. (Depends on whether it's loaded from pip in the testing setup.)

jpivarski:
A PR that only adds tests doesn't need a new version, but it's not a bad thing.

You didn't have to put the `nbytes` and `tostring` methods on all `TArray*` classes; putting them only on the superclass (`TArrray`) would be sufficient:

```python
    @property
    def nbytes(self):
        return len(self) * self._dtype.itemsize   # uses subclass's self._dtype

    def tostring(self):
        return numpy.asarray(self, dtype=self._dtype).tostring()   # uses subclass's self._dtype
```

reikdas:
> A PR that only adds tests doesn't need a new version, but it's not a bad thing.

I increased the version number because this PR also fixes the issue of not being able to create histograms from ROOT files read by uproot.

jpivarski:
Oh, I didn't see that in the changes. I only saw tests...

reikdas:
> Oh, I didn't see that in the changes. I only saw tests...

Those were the changes in the TArray classes in rootio.py (and the uproot_methods PR).

PR #266, opened 2019-03-29
Use mamba instead of conda for most builds

chrisburr:
Mamba is a prototype replacement for the conda environment solver, it can solve any environment in ~2 seconds instead of the minutes it takes conda. You can read more about it [here](https://medium.com/@wolfv/making-conda-fast-again-4da4debfb3b7).

Mamba an alpha and still has a couple of issues that cause it do behave differently to conda. I think it should be stable enough here with the extra constraints I've added and it makes the testing a lot faster but I understand if don't want to merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,24 +31,33 @@ env:
   - PYVER=pypy3.5 NPY="numpy>=1.15"
 
 install:
+  # Install conda
   - wget -nv http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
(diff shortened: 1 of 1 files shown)


jpivarski:
From my web searches, it does seem like the SAT solver is a common pain point. Talking with Jonathan at HOW, I learned what it's doing and why it's so resource-heavy. There are certainly situations where I'd rather have a working set of versions rather than the absolute best working set of versions.

I'm surprised, though, that this is a separate project and not a new option in conda. Maybe they'll get merged and it will be someday.

It could be okay if this tool gets a different set of packages than conda would. But for testing, we need it to always give the same set. And of course, it should also be a correct solution to the constraints.

jpivarski:
This works, and I can't think of what else needs to be checked to be sure it's safe. We'll use mamba until this feature becomes part of conda because it helps the build times so much.

PR #270, opened 2019-04-25
Make XRootDSource fork-safe

nsmith-:
By setting the appropriate environment variable.

The bug this fixes is a hang in child processes triggered by use of uproot with XRootD files in multiprocessing or concurrent.futures process pools after doing so in the parent process.  An example that triggers the bug is below:
```python
#!/usr/bin/env python
import concurrent.futures
import uproot

file = "root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/Run2012B_DoubleMuParked.root"

def read(fn):
    print("starting work")
    with uproot.open(fn) as f:
        print("opened file")
        sumw = f['Events'].numentries
    return sumw

print(read(file))

with concurrent.futures.ProcessPoolExecutor(max_workers=1) as executor:
    future = executor.submit(read, file)
    print(future.result())
```

### DIFF
--- a/uproot/source/xrootd.py
+++ b/uproot/source/xrootd.py
@@ -2,6 +2,7 @@
 
 # BSD 3-Clause License; see https://github.com/scikit-hep/uproot/blob/master/LICENSE
 
+import os
 import threading
 
 import numpy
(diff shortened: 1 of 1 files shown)


nsmith-:
I couldn't find much documentation on why one would ever _not_ want this option enabled.

jpivarski:
Unless you have any objection, I'll take this when the tests pass. It will combine with PR #269 to become uproot 0.5.1.

PR #271, opened 2019-04-30
Faster travis tests using mamba

reikdas:
If we skip the 2 tests that require Event.root (we were skipping those in appveyor already) in travis as well, mamba works. 
This PR adds speed to travis tests(by reverting to using mamba instead of conda) at the cost of skipping 2 tests. I am not sure if this is something we want, please close this PR if we do not. 

(Not related to this PR - Appveyor tests sometimes fail on any random job because of `Could not resolve host: github.com` while trying to git clone uproot.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,17 +37,23 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --add channels conda-forge;
+  - conda install --quiet --yes -c conda-forge/label/mamba-alpha mamba
   - conda config --set always_yes yes --set changeps1 no
   # Create the conda testing environment
+  # FIXME: Mamba decides to upgrade Python here so pin it again
   # FIXME: Channel priority is also mixed up
+  # FIXME: Mamba doesn't install pip by default
+  # FIXME: Mamba causes pip install numpy to be extremely slow
   - if [[ "${PYVER}" = pypy* ]]; then
       conda create --quiet --yes -n testenv ${PYVER};
+    elif [ "${PYVER}" = "2.7" ] || [ "${PYVER}" = "3.6" ] || [ "${PYVER}" = "3.7" ]; then
+      mamba create --quiet --yes -n testenv python=${PYVER} pip;
     else
       conda create --quiet --yes -n testenv python=${PYVER};
     fi
   - source activate testenv
   - if [ "${PYVER}" = "2.7" ] || [ "${PYVER}" = "3.6" ] || [ "${PYVER}" = "3.7" ]; then
-      conda install --quiet --yes python=${PYVER} root;
+      mamba install --quiet --yes python=${PYVER} pip root;
       source activate testenv;
     fi
   - pip install --upgrade setuptools-scm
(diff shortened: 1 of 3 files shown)


jpivarski:
Unfortunately, one of these two is part of a work in progress by @plexoos (adding missing branch interpretations), so I wanted to keep them in _some_ test. They have to be accessed remotely because the Event.root file is just a little too large to be in a GitHub repository (just over a sanity-check limit).

We could, perhaps, `wget` the file in `.travis` and then access it as a local file in these two tests. I don't know if `wget` exists on AppVeyor.

jpivarski:
Also, it's extremely weird that conda vs mamba, which has to do with installing prerequisites, would be incompatible with an HTTP request in one of the subsequent tests. Maybe each test has a network usage budget? Constraints between seemingly unrelated components are things to avoid whenever possible (though this one is out of our hands).

reikdas:
> We could, perhaps, wget the file in .travis and then access it as a local file in these two tests.

We would need to find a way to have the file be present locally(or a way to access the file over the network without travis) as well, otherwise, if a dev ran the tests manually on their local machine, it would fail due to the file being absent.

jpivarski:
That's an excellent point. What I was suggesting would have created a long-distance dependency like the one I was just complaining about! I think that pytest has some kind of `skipif` where you get to put in a predicate. The predicate can be `os.path.exists(...)`.

plexoos:
As an alternative, we could also reduce the size of Even.root by saving fewer events. I believe it contains 100 events now but 10 events might be enough for the current tests in which we just read a few hand-picked values from the tree.

jpivarski:
@plexoos If you can do that, then it would simplify matters. Event.root could be moved back into tests/samples, and ultimately scikit-hep-testdata. I have an outstanding PR for getting everything from scikit-hep-testdata, but something wasn't working and handling Event.root as a special case complicated things.

plexoos:
Ok, I can look into it. But I still wonder why would there be a problem with cloning scikit-hep-testdata if your CI service allows you to clone this repository...

jpivarski:
The issue with scikit-hep-testdata (PR #237) wasn't cloning, it was that it wasn't pip-installable due to some directories trickery they tried (scikit-hep/scikit-hep-testdata#14). I guess I could have cloned scikit-hep-testdata and the tests would have passed, but that would be a manual step. The tests should automatically download their dependencies through `tests_require` and that requires pip-installability.

Anyway, the issue with scikit-hep-testdata is independent of Events.root being too large. We had two solutions for that: (1) wget Event.root with a `pytest.skipif` on the tests that use it, or (2) make Event.root smaller and treat it like any other test sample. If the latter, then eventually it will make its way into scikit-hep-testdata when that works, which would be a bonus.

reikdas:
The current solution does not skip the tests, it downloads the file separately if on travis or appveyor.

jpivarski:
I'll take a look at this and maybe include it in uproot 3.6. However, this is less urgent because it only changes the testing procedure, not the library itself.

Thanks, though! (Faster tests may be an important means to the end of cleaning up the library for versino 3.6.)

PR #278, opened 2019-05-20
Awkward lazyarrays

jpivarski:
This will become a new semi-major version, uproot 3.6, and will be coordinated with uproot-methods 0.6 (including scikit-hep/uproot-methods#55). For good measure, there may be a new semi-major version of awkward as well.

@reikdas, do you want me to include #258 in that as well? I'm gathering up features for a presentation next Tuesday, so they need to be robust.


reikdas:
> @reikdas, do you want me to include #258 in that as well? I'm gathering up features for a presentation next Tuesday, so they need to be robust.

#258 is working. If you could take a look and check if the code design is okay, I think it is ready to be included.

jpivarski:
This passed the tests, though I cancelled all but one early. (The N-1 run was indicative.)

PR #282, opened 2019-06-02
uproot can now write TProfile and TProfile2D

reikdas:
This is still a WIP because test test_issue57 fails. 
Test test_issue57 does not have any relation to ROOT file writing. The commits in this PR do not contain any changes to the reading code or the testing procedure so it is not very clear to me why that test fails here but passed in master when they were pushed - I just created a branch off master and restarted the tests and they failed(It is a NameError so it should not be dependent on something going wrong in the Appveyor or Travis builds). Strangely, the test passes on my local machine. 
(This PR contains a lot of commit logs according to the Github diff that are already in master because I created this branch from a branch that has already been merged into master)


jpivarski:
This problem with the `NameError` in `test_issue57` does _not_ make sense. It does not make sense that it works locally (for me, too) and not on Travis without rewriting the generator comprehension, and it does not make sense that rewriting the generator comprehension should fix it.

However, it was a rather complex generator comprehension (two levels deep) and pytest works by rewriting code. It is possible that some version of pytest did that incorrectly for this corner case in the Python syntax, that version was on Travis and a fixed version of pytest was on our laptops. If so, that's a little disturbing.

(At least visually, it reminded me of Fortran errors—old versions of Fortran would simply stop interpreting your source code past column 70 or whatever, so if you were lucky, your expression wouldn't make sense when cut off at column 70 and raise an error. If you were unlucky, like if the cut-off was right _before_ the `+` of the last term in an expression, then it would run without telling you it was dropping the last term. This bug _looked_ like that because it wasn't seeing that the name `y` does exist and is defined near the end of the line...)

Anyway—works now. Have fun!

jpivarski:
@reikdas Next time, bump the version number so that I can deploy a release with the new features.  :)

reikdas:
Sorry, I keep forgetting to do that :(

PR #288, opened 2019-06-09
Writing LZ4 compressed ROOT files

### LINKED ISSUE #279: Write LZ4 compressed files
uproot is currently able to write ZLIB and LZMA compressed ROOT files, but not LZ4 compressed ROOT files.

reikdas:
Fixes #279

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -143,7 +143,7 @@ setup(name = "uproot",
       test_suite = "tests",
       install_requires = ["numpy>=1.13.1", "awkward>=0.10.0", "uproot-methods>=0.6.0", "cachetools"],
       setup_requires = ["pytest-runner"],
-      tests_require = ["pytest>=3.9", "pkgconfig", "lz4", 'backports.lzma;python_version<"3.3"', "mock", "requests"],
+      tests_require = ["pytest>=3.9", "pkgconfig", "lz4", 'backports.lzma;python_version<"3.3"', "xxhash", "mock", "requests"],
       classifiers = [
           "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -143,7 +143,7 @@ def get_description():
       test_suite = "tests",
       install_requires = ["numpy>=1.13.1", "awkward>=0.10.0", "uproot-methods>=0.6.0", "cachetools"],
```
nsmith-: Shouldn't `xxhash` appear here rather than `tests_require` ?

jpivarski: For pip, which is determined by setup.py, lz4 and therefore xxhash are optional dependencies, though we want to test them. For conda, they're strict dependencies, and that's controlled by uproot-feedstock.

nsmith-: hmm, well then [this import](https://github.com/scikit-hep/uproot/blob/master/uproot/source/compressed.py#L11) should be wrapped by a `try:` statement if it is optional.  I encountered an error because I didn't have `xxhash` installed.

jpivarski: Yes, it absolutely must. @reikdas?

jpivarski: Actually, it should be inside of the function call, just like `import lz4` is, and probably right after `import lz4` to indicate the relationship. Although I think `xxhash` is used in a different function... anyway, it should be inside of a function so that it doesn't load on startup if you have it but don't use it.

reikdas: Sorry, this was my mistake. I opened a PR fixing this. Also @jpivarski I have a draft release ready once you have merged it.

PR #292, opened 2019-06-22
Use curl -L to have trigger_binder follow URL redirect

matthewfeickert:
# Description

BinderHub now operates as a Binder Federation and so there is no guarantee where the Binder build will take place. As a result, the Binder trigger must follow any redirect of the URL.

c.f.
https://discourse.jupyter.org/t/problem-triggering-binder-build-through-api-endpoint/1440

### DIFF
--- a/binder/trigger_binder.sh
+++ b/binder/trigger_binder.sh
@@ -3,7 +3,7 @@
 function trigger_binder() {
     local URL="${1}"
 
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@jpivarski This is ready for review. This is a clone of [pyhf PR 487](https://github.com/diana-hep/pyhf/pull/487).

PR #295, opened 2019-06-25
Add from __future__ import absolute_import

### LINKED ISSUE #294: Name clash between pandas.py & pandas (python 2 on SWAN)
I am using uproot on CERNs SWAN, which is hosting a jupyter notebook with a python 2 kernel. uproot 3.7.0 has been installed via `pip install --user uproot` on the SWAN node, which is essentially putting the python module into a temporary home dir on the network storage system EOS.

Now copy pasting a section from uproots examples in binder
```python
import uproot
for df in uproot.pandas.iterate("http://scikit-hep.org/uproot/examples/Zmumu.root", "events", "p[xyz]1", entrysteps=500):
    print(df[:3])
```
into the notebook, resulted in the following error:
```
AttributeError: 'module' object has no attribute 'DataFrame'
```
Luckily the Internet is big enough, that just throwing the error message into google lead me to a [useful stack overflow discussion](https://stackoverflow.com/questions/20621607/module-object-has-no-attribute-dataframe), and renaming `~/.local/lib/python2.7/site-packages/uproot/pandas.py` to something else and replacing in `__init__.py`
```python
from uproot import pandas
```
with
```python
from uproot import somethingelse as pandas
```
really solved this problem.

I am not sure if this is just some kind of edge case or if I'm doing something silly without knowing it, but I thought I'd raise this, since other uproot users might run into this as well.

edit:
Ah, I just thought of something else.
In order to load the users python modules on SWAN, you have to start the instance with a script that modifies the environment like this:
```bash
#! /bin/env sh
export PYTHONPATH=$CERNBOX_HOME/.local/lib/python2.7/site-packages:$PYTHONPATH
```
is this bad practice in some sense?

chrisburr:
Fixes #294, if you'd rather do it differently feel free to close this.


jpivarski:
I didn't finish reading my emails and didn't see that you did this. I guess it's another way to solve it (see the issue for my thoughts). The advantage of this one is that absolutely nothing changes. I'll think about it.

PR #297, opened 2019-06-26
Some Taxis attributes

reikdas:
The only TAxis attributes left now are fLabel and fModLabel. (I didn't bump the version number because I am not sure if anyone really uses these attributes and if it these changes would justify a new release)


uproot/write/objects/TH.py
```diff
@@ -270,7 +270,7 @@ def length_tattaxis(self):
     def return_taxis(self, cursor, axis):
         cnt = numpy.int64(self.length_taxis(axis) - 4) | uproot.const.kByteCountMask
         vers = 10
-        if axis["_fFirst"] != 0 or axis["_fLast"] != 0 or axis["_fBits2"] != 0 or axis["_fTimeDisplay"] or axis["_fTimeFormat"] != b"" or axis["_fLabels"] or axis["_fModLabs"]:
+        if axis["_fLabels"] or axis["_fModLabs"]:
```
jpivarski: I prefer not to rely on Python's "truthiness" in boolean expressions. So many values can resolve to true or false that I'd much rather have something like

```python
if axis["_fLabels"] is not None or axis["_fModLabs"] is not None:
```

or even better for this case, since I know that only a minimum of fields should be required of these objects:

```python
if axis.get("_fLabels", None) is not None or axis.get("_fModLabs", None) is not None:
```

so that missing keys are treated the same way as `None` keys.

uproot/write/objects/TH.py
```diff
@@ -206,7 +206,7 @@ def return_tlist(self, cursor, values):
         for value in values:
             raise NotImplementedError
         return (cursor.return_fields(self._format_cntvers, cnt, vers) + self.return_tobject(cursor) +
-            cursor.return_string(b"") + cursor.return_fields(self._format_tlist, len(values)))
+            cursor.return_string(b"") + cursor.return_fields(self._format_tlist, len(values))) # Need to fix
```
jpivarski: The spelling is usually `FIXME`, which makes it easier to find these by an incremental search.

uproot/write/objects/TH.py
reikdas: I replaced the line with this - 
`if axis.get("_fLabels", None) != [] or axis.get("_fModLabs", None) != []:`

uproot/write/objects/TH.py
```diff
@@ -270,7 +270,7 @@ def length_tattaxis(self):
     def return_taxis(self, cursor, axis):
         cnt = numpy.int64(self.length_taxis(axis) - 4) | uproot.const.kByteCountMask
         vers = 10
-        if axis["_fLabels"] or axis["_fModLabs"]:
+        if axis.get("_fLabels", None) != [] or axis.get("_fModLabs", None) != []:
```
jpivarski: Wait a minute—if the null condition is that `axis["_fLabels"] == []` (same for `_fModLabs`), then you need to do

```python
if axis.get("_fLabels, []) != [] or axis.get("_fModLabs", []) != []
```

and if you want to be agnostic about whether it's a list, a tuple, or something else, you can do

```python
if len(axis.get("_fLabels", [])) > 0 or len(axis.get("_fModLabs", [])) > 0
```

Be sure you understand what `axis.get` is doing before using it in code.

jpivarski:
If this passes tests, you can merge it.

reikdas:
> If this passes tests, you can merge it.

I am not authorized to merge it.

jpivarski:
Oh, okay.

PR #298, opened 2019-06-26
Added test to check if directory size is increased

reikdas:


jpivarski:
If the tests pass, merge this one, too.

jpivarski:
(The only reason they might not is because I merged a conflict.)

PR #299, opened 2019-06-27
Trigger Binder builds on both GKE and OVH clusters

matthewfeickert:
# Description

Amends PR #292 

Given that the Binder Federation exists on both GKE and OVH there is no guarantee that the cluster that a build is triggered on is the same cluster that a user will be directed to when they launch the Binder. As a result, both clusters should be triggered when a PR is merged.

c.f. https://discourse.jupyter.org/t/the-binder-federation/1286

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,5 +109,6 @@ jobs:
     before_install: skip
     install: skip
     script:
-    # Use Binder build API to trigger repo2docker to build image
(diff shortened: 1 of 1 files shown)


jpivarski:
Oh, okay. I'll merge this before the tests pass because this change doesn't affect the tests that run in a PR but the test that comes after the PR merge. I'll watch it, though, to be sure nothing breaks.

PR #300, opened 2019-06-28
Test ROOT versions 6.16.00 and 6.18.00

jpivarski:
Making sample files with ROOT:

```bash
$ cd tests/samples
$ root -l
root [0] .L sample4version.C++
root [1] sample4version("6.16.00", "uncompressed")
```

for each of `"uncompressed"`, `"zlib"`, `"lzma"`, and `"lz4"`.


jpivarski:
@reikdas Please do this for ROOT 6.18!

jpivarski:
Same deal with this one: if the tests pass, merge it. (I'm basically saying that I've reviewed it. Maybe we should get into a pattern where you request a review when you're ready and I just push the "Approve" button without lengthy explanations. The "Approve" would be a signal that you're allowed to merge.)

PR #301, opened 2019-06-28
ROOT 6.18.00 streamers

reikdas:


jpivarski:
I've reviewed the differences. If this passes tests, merge it. (If you don't have merge access, let's change that, though I'd still prefer to manually scan the diffs and only give you permission to merge if I don't see any issues.)

PR #305, opened 2019-07-16
Bin labels can now be written

### LINKED ISSUE #132: Write custom fLabels
Each axis has a `fLabels` member for labeling the axis with strings, rather than numbers. Physicists do use this regularly. Currently, `fLabels` is in the short block that we "punted" on, raising a `NotImplementedError` if `fLabels` is not empty.

File `tests/samples/issue33.root` has an example of a histogram with labels. Use this as an example to handle `fLabels` properly.

reikdas:
Fixes #132

### DIFF
--- a/uproot/write/objects/TH.py
+++ b/uproot/write/objects/TH.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import
 
 import struct
+from copy import copy
 
 import numpy
 
@@ -82,6 +83,9 @@ class TH(object):
         self.fields["_fSumw2"] = numpy.array(self.fields["_fSumw2"], dtype=">f8", copy=False)
         self.fields["_fBinEntries"] = numpy.array(self.fields["_fBinEntries"], dtype=">f8", copy=False)
         self.fields["_fBinSumw2"] = numpy.array(self.fields["_fBinSumw2"], dtype=">f8", copy=False)
+
+        #For _writeobjany
+        self._written = {}
 
     @staticmethod
     def fixstring(string):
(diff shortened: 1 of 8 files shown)


uproot/write/objects/TH.py
```diff
@@ -83,6 +84,9 @@ def __init__(self, histogram):
         self.fields["_fBinEntries"] = numpy.array(self.fields["_fBinEntries"], dtype=">f8", copy=False)
         self.fields["_fBinSumw2"] = numpy.array(self.fields["_fBinSumw2"], dtype=">f8", copy=False)
 
+        #For _writeobjany
+        self._written = {}
```
jpivarski: Is it absolutely necessary for the object to know that it has been written? We have a model in which

```python
file["something"] = histogram
```

makes a copy of the `histogram` that is unconnected to any other `histogram`—for instance,

```python
file["something_else"] = histogram
```

makes an entirely new, unconnected object in the file. For one thing, that file object has a different name, while the in-memory object has no name.

If `_writeobjany` wants to keep references to objects that have been written so that it doesn't write the same thing twice, well, we're writing the same thing twice to stay consistent with our "by value" rather than "by reference" model.

... Unless you can convince me otherwise that this model doesn't apply here?

uproot/write/sink/cursor.py
```diff
@@ -21,6 +21,7 @@ def write_fields(self, sink, format, *args):
         self.index += format.size
 
     def return_fields(self, format, *args):
```
jpivarski: All of these methods with `return_` in their names are doing more than just returning a value without side-effects, they're updating the position of the cursor, so it might be better to say `write_` instead of `return_` or some other word that suggests an in-place update.

I know that the cursor doesn't literally write data to a file, which is an argument against using the word `write_`. You have `update_` meaning a different thing below. Maybe the prefix should be `put_`? "Put" connotes a change in state.

uproot/write/objects/TH.py
reikdas: I completely overlooked this - the histogram object should not know that it has been written. (This would cause problems as soon as I started working on TTrees anyway)

But, there should be a dictionary somewhere holding the class names (and their corresponding indices) that have already been written. (Side note for future reference - the dictionary does not write the objects that the user wants to write the file - for eg. histograms, but it instead writes constituent parts such as TList and THashList).

I feel like moving the writeobjany (or as it is now called put_objany) to a different file/class and having the dictionary there as well, would be the correct course of action and I am currently trying to incorporate that.

reikdas:
No I think that should be in the same PR.. considering that it is basic functionality one would expect to come with this feature.

jpivarski:
Okay. When you have passing tests of writing several histograms with fLabels to the same file, go ahead and merge the PR. It's looking good.

Remember that the most important thing that will come of this is a thorough understanding of `put_objany` so that TTree-writing can proceed with fewer uncertainties.

reikdas:
There is one case of writeobjany that I have not yet implemented because I have been unable to think of any test case to trigger it.. but once I do encounter it, it will be fairly easy to incorporate because the implementation is very similar to the already existing cases.

PR #307, opened 2019-07-24
Comparing checksum while reading lz4 compressed files

reikdas:


uproot/source/compressed.py
```diff
@@ -127,8 +130,17 @@ def _prepare(self):
...
                     compressedbytes -= 8
+                    checksum = cursor.bytes(self._compressed, 8).tolist()
+                    copy_cursor = copy(cursor)
+                    after_compressed = copy_cursor.bytes(self._compressed, compressedbytes)
+                    if sys.version_info < (3,):
+                        def iterbytes(data):
+                            return (ord (char) for char in data)
+                    else:
+                        iterbytes = lambda x: iter(x)
+                    if [x for x in iterbytes(xxhash.xxh64(after_compressed).digest())] != checksum:
```
jpivarski: Does this iterate over all (compressed) bytes in the basket with a Python for loop? Maybe that's not an error but it would be a horrendous performance bug (on the compression method chosen for speed). Same for line 134 above: `numpy.ndarray.tolist` and `for x in numpy.ndarray` are not allowed (same for writing, when the dataset is large enough).

`xxhash` must have an interface that takes a whole array or a whole bytestring or a whole memoryview at a time. The number of steps in Python must not scale with the size of the data to compress or uncompress.

jpivarski:
You also need to bump the version number. Thanks!

reikdas:
> You also need to bump the version number. Thanks!

There was no release for the last version number bump. Should we still bump the version number?

jpivarski:
I must have forgotten it. Yes, 3.8.1 has not been released so 3.8.1 should be the next release; no need to bump this one.

You'll need to help me remember what went into 3.8.1 so that I can write the release notes. If you have permission to _draft_ a release, please do so, and I'll press the "publish" button.

uproot/source/compressed.py
```diff
@@ -130,10 +130,10 @@ def _prepare(self):
                 elif algo == b"L4":
                     compression = self.compression.copy(uproot.const.kLZ4)
                     compressedbytes -= 8
-                    checksum = cursor.bytes(self._compressed, 8)
+                    checksum = cursor.field(self._compressed, struct.Struct(">Q"))
```
jpivarski: Good, though generally we predefine the `struct.Structs` (as `_fields0`, etc.).

You can probably reuse a lot of these techniques when writing LZ4 checksums.

reikdas: > Good, though generally we predefine the struct.Structs (as _fields0, etc.).

Should I make that change? 

> You can probably reuse a lot of these techniques when writing LZ4 checksums.

I don't think so because in the case of writing there are no conversions. We just have a bunch of bytes, put them into xxhash to get the checksum bytes and put them into the file.

PR #315, opened 2019-08-14
Fixes fSeekFree

reikdas:
ROOT can now write objects in files created by uproot.


jpivarski:
Wow, that was simple. After I've merged this, do a merge-master to get it into your working branch.

PR #332, opened 2019-09-05
Don't attempt to set a pandas.RangeIndex in-place.

jpivarski:


jpivarski:
A solution for Pandas before and after Pandas version 0.25 is in PR #351, which will eventually be deployed as uproot version [3.10.2](https://github.com/scikit-hep/uproot/releases/tag/3.10.2).

PR #342, opened 2019-09-20
Fixed extend for flush=True

reikdas:
Fixes extend with flush=True (even though it is still using the inefficient implementation of flush=False).
Even after the better(and correct) implementation of flush=False for extend, the code under the flush=True block should probably not change.

Although, in its current implementation this makes extend=True less efficient, the implementation is now correct.


jpivarski:
Ah—this is the other PR.

PR #346, opened 2019-09-24
Fix flushsize default

reikdas:


reikdas:
@jpivarski I have changed the default flushsize to 20 MB. The closest I checked was 50 MB - which failed.
Should I merge this or look for the approximate point where travis breaks because of MemoryError?

jpivarski:
It's not necessary to find the exact point where Travis fails (that would likely change in the future, anyway). I would have jumped in powers of 10 down to 10 MB, but if 20 MB works, let's go with it.

reikdas:
I tried multiples of 10 - 30 MB seems most appropriate. I merged this into master and drafted a release.

PR #353, opened 2019-09-28
Raise NotImplementedError for open issues we plan to handle

reikdas:
.. and conversion of basket data to a numpy array of the required data type in extend.

### DIFF
--- a/uproot/version.py
+++ b/uproot/version.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 
 import re
 
-__version__ = "3.10.2"
+__version__ = "3.10.3"
 version = __version__
 version_info = tuple(re.split(r"[-\.]", __version__))
 
--- a/uproot/write/compress.py
+++ b/uproot/write/compress.py
@@ -67,6 +67,10 @@ def write(context, cursor, givenbytes, compression, key, keycursor):
 
     _header = struct.Struct("2sBBBBBBB")
     uncompressedbytes = len(givenbytes)
+
+    if uncompressedbytes > 2**24:
+        raise NotImplementedError("Data larger than 2^24 bytes cannot be compressed yet")
+
     u1 = (uncompressedbytes >> 0) & 0xff
     u2 = (uncompressedbytes >> 8) & 0xff
     u3 = (uncompressedbytes >> 16) & 0xff
--- a/uproot/write/objects/TTree.py
+++ b/uproot/write/objects/TTree.py
@@ -103,6 +103,10 @@ class TTree(object):
         for key, value in branchdict.items():
             branchdict[key] = numpy.array(value, dtype=self._branches[key]._branch.type, copy=False)
 
+            #Temporary place holder
+            if branchdict[key].ndim != 1:
+                raise NotImplementedError("Multi dimensional array support is coming soon")
+
         #Check for equal number of values in baskets
         values = iter(branchdict.values())
         first = next(values)
--- a/uproot/write/objects/TTree.py
+++ b/uproot/write/objects/TTree.py
@@ -340,6 +340,10 @@ class TBranch(object):
         self._branch.fields["_fBasketEntry"][self._branch.fields["_fWriteBasket"]] = self._branch.fields["_fEntries"]
         self._branch.fields["_fEntryNumber"] += len(items)
         basketdata = numpy.array(items, dtype=self._branch.type, copy=False)
+
+        if basketdata.ndim != 1:
+            raise NotImplementedError("Multi dimensional array support is coming soon")
+
         givenbytes = basketdata.tostring()
         cursor = uproot.write.sink.cursor.Cursor(self._branch.file._fSeekFree)
         self._branch.fields["_fBasketSeek"][self._branch.fields["_fWriteBasket"] - 1] = cursor.index
--- a/uproot/write/objects/TTree.py
+++ b/uproot/write/objects/TTree.py
@@ -99,6 +99,10 @@ class TTree(object):
         self._tree.write(context, cursor, name, key, copy(keycursor), util)
 
     def extend(self, branchdict, flush=True):
+        #Convert to numpy arrays of required dtype
+        for key, value in branchdict.items():
+            branchdict[key] = numpy.array(value, dtype=self._branches[key]._branch.type, copy=False)
+
         #Check for equal number of values in baskets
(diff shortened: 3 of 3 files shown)


jpivarski:
Did you verify that these NotImplementedErrors get triggered in the cases the users are complaining about? Does it prevent the "corrupt files"? (I think they have steps to reproduce—raising an error in those cases would be a step up.)

reikdas:
Yes, I have verified that the `NotImplementedError`s get triggered when trying to compress data larger than 2**24 bytes or if a user tries to write multi-dimensional arrays in a basket.

jpivarski:
Okay, good. If you haven't already, you can tell the people who have been affected that there's at least this protection for now, but keep those issues open as a reminder that they need to be covered.

Actually, the `flushsize="30 MB"` puts us in danger of hitting that frequently for small numbers of branches—it's probably one of the first that should be fixed. (Particularly because it's the most straightforward—a one-time write gets replaced by a `while` loop.)

reikdas:
@jpivarski Should we merge this? (and create a release)

jpivarski:
Yes, merge and release this to focus on the ROOT bug. (Be sure that the NotImplementedError for too-large baskets suggests writing smaller baskets, because with the 1 MB flush size, a lot of users would hit that. It's only text in the error message—no need for a full Travis cycle.)

reikdas:
> because with the 1 MB flush size

Isn't the flush size 30 MB?

> Actually, the flushsize="30 MB" puts us in danger of hitting that frequently for small numbers of branches—it's probably one of the first that should be fixed. 

Won't the flushsize be hit more often if there are more branches? (Because if the flushsize specifies the size in bytes (B/KB/MB/GB), then it is divided equally among the branches?)

> (Particularly because it's the most straightforward—a one-time write gets replaced by a while loop.)

I don't understand :( - If the flush size is reached, the basket data is written to the file and the branch buffer is emptied, this is implemented in master.

jpivarski:
The 30 MB (you're right) is divided among all the branches, so if there are fewer branches there will be larger baskets. Anyway, I'm just saying that the error message should suggested users how they can work around it, namely by writing smaller baskets. And this is more or less moot well we're not claiming that writing works yet. So just merge this and move on.

reikdas:
I have drafted a release.

reikdas:
> The 30 MB (you're right) is divided among all the branches, so if there are fewer branches there will be larger baskets.

But if there are fewer branches, although the baskets will be larger, the flushsize will also be larger which would balance it out? 
In any case, when should we throw this error? - because the code would still work although inefficiently if the flushsize is not correctly set. (We would not know the number of events beforehand, so we cannot perform some rough calculation and let the user know if their given flushsize would cause serious issues)

jpivarski:
I'm beginning to wonder if buffered writes are worth the complexity. We could just have the `extend` method require all arrays in its dict be equal-length and immediately call `newbasket` on each one of them (no `flushsize`, no `flushfactor`, no `flush=False`). Then `extend` wouldn't be doing any buffering, just sanity checking to ensure that all branches have the same number of entries. That, by itself, is a benefit, a reason to use `extend` over `newbasket`.

The usual benefit of buffered writing is that writing has high latency—it takes a long time to start the writing process and wait for it to finish—which would be a disaster for an interface that takes a few bytes at a time, such as `fwrite` (the C function that writes bytes to a file). ROOT is in this situation, since its interface takes one entry at a time, which can be very small—it needs to be buffered. But our interface almost always takes arrays at a time. As long as the arrays are reasonably large, buffers are not needed from a performance point of view. We just wanted to decouple the basket size from the array size, but if we're having all this trouble creating an interface to understand the user's intended basket size, maybe we're making things unnecessarily complicated.

If we have no buffering ever—the `extend` method just checks arrays for sanity and makes each a new basket—then we put basket sizes in the user's hands, implicitly through their array size. It will become a documentation issue, we'll have to carefully explain that small arrays → small baskets → bad. In that world, the `append` method (which never made _much_ sense) makes _no_ sense and should not be included.

What do you think? Scaling back on complexity is good, particularly if it introduces not-originally-foreseen problems.

reikdas:
I think selective flushing is nice to have.

The only issue I see with the intended flushing interface is if the user does not increase the flushsize themselves if they have a large number of branches. This can be solved with a larger default flushsize(we have settled on 30 MB because of Travis, we could easily increase this to even 100 MB by scaling down the Travis tests and I don't think any user would complain with the additional space being taken up by the ROOT files written by uproot). Thoughts?
Also, once the proper flushing mechanism is in place, good documentation both in the README and the readthedocs about how to use the flushing mechanism and how it works (something we can point to users posting issues about large file sizes) should suffice.

What we _can_ do is probably implement - 

> We could just have the extend method require all arrays in its dict be equal-length and immediately call newbasket on each one of them (no flushsize, no flushfactor, no flush=False). Then extend wouldn't be doing any buffering, just sanity checking to ensure that all branches have the same number of entries. 

till I get the proper flushing mechanism in place.

jpivarski:
Of course no decision is final—deciding to not put flushing in now doesn't mean we can't decide to add it later. I've gone back and forth on a lot of features, and git is great because I can easily remind myself how it was done—nothing is truly lost.

What you said in the last paragraph, remove it for the time being, is exactly what I'm saying. There are enough troubles—let's remove it until we know we need it.

Have you heard of [You ain't gonna need it](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)? We don't know yet from the users that it's important. If it were trivial to implement, we might put it in as one of the "bells and whistles" to see if anybody finds it useful. (The `TTree.keys/values/items` methods have a bunch of "bells and whistles" filtering parameters that I didn't think anybody was going to use, and first heard feedback about a year and a half after writing them. If it was difficult to implement them, it would not have been a good use of time, but they were simple to write and easy to hide from users who didn't care.)

I first assumed we should have buffering because it's normal for writing interfaces to have that. But most writing interfaces take many small inputs, so they absolutely need buffering for performance. This really isn't true of our interface—uproot is array-oriented (in both reading and writing).

How about we put this out there with no flushing (or at least make the current implementation inaccessible), explain "array size = basket size" in the documentation, and say that if making appropriate array sizes manually is a hardship, we'll revisit flushing upon request. Then we can be more Agile; directing programming effort to the problems that really matter to users by first finding out what really matters to users.

Issue #352, opened 2019-09-27
Reading a file written with uproot: TTreeCache::{AddBranch,DropBranch}: unknown branch -> *


douglasdavis:
I'm taking a first stab at TTree writing and I'm running into an issue when trying to read the file back into ROOT (reading it back in uproot is working fine). ROOT is telling me
```
Error in <TTreeCache::AddBranch>: unknown branch -> *
Error in <TTreeCache::DropBranch>: unknown branch -> *
```
when I try to call `TTree::Show` on it. It seems like some kind of nameless ghost of a branch is appearing in the tree some how. I've also tested reading the uproot-originating tree via `RDataFrame` (to create ROOT histograms with `RDataFrame::Histo1D`). Things just crash.

My file is created like so:

```python
## `out_dict` is a dictionary of {name(type=str) -> data(type=numpy.ndarray)}.
## `tree` is just a string for the tree name in the file.

branches = {}
for colname, arr in out_dict.items():
    if arr.dtype == np.uint32: ## convert to int32 for now
        out_dict[colname] = out_dict[colname].astype(np.int32)
    branches[colname] = uproot.newbranch(out_dict[colname].dtype)
    log.info(f"Saving branch {colname} with dtype {out_dict[colname].dtype}")

with uproot.recreate(outfile) as f:
    f[tree] = uproot.newtree(branches, flushsize="10 MB")
    f[tree].extend(out_dict, flush=False)
```

I've tried tinkering with the flush (not declaring a `flushsize` on the tree and giving extend `flush=True`) and compression settings just to see if it might impact the error I'm seeing but it doesn't (just file size is different, as I guess is expected). [Link to file](http://webhome.phy.duke.edu/~ddavis/public/uproot352.root)

douglasdavis:
Looks like we can get a branch named `*` [through logic on the ROOT side](https://github.com/root-project/root/blob/master/tree/tree/src/TTreeCache.cxx#L519-L523), but I can't quite follow how this is triggered

reikdas:
Thanks for reporting the issue!

> `## `out_dict` is a dictionary of {name(type=str) -> data(type=numpy.ndarray)}.`

To aid in debugging, could you tell me how many dimensions does data have? (`data.ndim`)

jpivarski:
[...]

PR #357, opened 2019-09-29
Remove TTree writing documentation till #352 is fixed.

reikdas:


reikdas:
> The only thing that's lacking is to also comment out the table of contents item.

Done. Should I merge?

jpivarski:
Yes, and you don't need to wait for Travis because it's just a readme change.

PR #362, opened 2019-10-02
Update documentation

reikdas:


README.rst
```diff
@@ -4204,9 +4204,7 @@ ecosystem, are just a tuple of counts/bin contents and edge positions.
 Creating and writing data to ROOT files
 =======================================
 
-Uproot has a limited (but growing!) ability to *write* ROOT files. Two
-types currently supported are ``TObjString`` (for debugging) and
-histograms.
+Uproot has a limited ability to *write* ROOT files. Three types currently supported are ``TObjString`` (for debugging), histograms and TTrees.
```
jpivarski: Now that we have TTrees, let's lead with that:

    Uproot has a limited ability to *write* ROOT files, including TTrees of flat data (non-jagged:
    single number per event), a variety of histogram types, and ``TObjString`` (for metadata).

PR #368, opened 2019-10-04
Resize all other branches when one branch overflows

### LINKED ISSUE #345: root file, written with uproot, size issue
Hi,

it seems that files written with uproot have a very large size. I tried to write a subset of a dataset (*less than 1%*) with less variables to another root file and the size of the latter is of the same order of magnitude as the size of the original root file.

I tried to play by adding a compression as described in the README

```python
tree = uproot.newtree(branchdict, compression=uproot.LZ4(4))
```

and also to reduce the data type of the branches from float64 to float32. This does reduce the size of the output root file but it is still very large.

reikdas:
Fixes #345

### DIFF
--- a/uproot/version.py
+++ b/uproot/version.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 
 import re
 
-__version__ = "3.10.4"
+__version__ = "3.10.5"
 version = __version__
 version_info = tuple(re.split(r"[-\.]", __version__))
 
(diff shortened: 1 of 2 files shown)


reikdas:
> Does this demonstrably reduce the number of TTree-metadata rewrites when the number of branches is large? Have you tested it?

Yes. In my tests, it reduces the size of the file from 10+ GB(I stopped the program execution at that point) to 4.5 GB.

jpivarski:
Okay, good. (This is the important thing.)

I restarted the one PyPy job that failed; it was probably an error in downloading the dependencies or something.

PR #369, opened 2019-10-05
Removing the flushing interface

reikdas:
Maybe the documentation in the README has not been correctly written.. @jpivarski could you please edit it if required?


reikdas:
I can use the branch I was intending to use for implementing the flushing interface to do that - https://github.com/scikit-hep/uproot/tree/ttree-write-flush 
Thanks for the suggestion!

PR #375, opened 2019-10-11
Optionally disable chunking in lazyarrays

nsmith-:
Only for the lazyarrays method of Tree and Branch, because for any larger scope chunking is necessary!


jpivarski:
I like the idea. There's an `entrysteps` parameter, which should give you an unchunked array is it's equal to `numpy.inf`—how does that interact with this? What if there are multiple files (in `uproot.lazyarrays`)? I think what it does is it gives a `ChunkedArray` with a single chunk. It could be better to drop the `ChunkedArray` if there's only ever going to be a single chunk.

jpivarski:
This might also interact badly with profiles, which are in uproot-methods, necessitating a multi repo version alignment. The profile code, such as the CMS nano, assumes a structure. It's going to assume that the ChunkedArray layer is there even if it's only one chunk.

nsmith-:
Originally I thought to remove the chunk wrapper if `entrysteps == float('inf')` but there's a valid case where that might happen when using the multi-file `lazyarrays` method, in which case the existing behavior of creating one chunk per file would be expected.
As for profiles, the default behavior is compatible, so we can migrate uproot-methods at our leisure.

uproot/tree.py
```diff
@@ -546,8 +546,10 @@ def wait():
     def lazyarray(self, branch, interpretation=None, entrysteps=None, entrystart=None, entrystop=None, flatten=False, awkwardlib=None, cache=None, basketcache=None, keycache=None, executor=None, persistvirtual=False):
```
jpivarski: (I'm looking at this on a laptop, now.) The `TTree.lazyarray` method should have and pass on a `chunked=True` parameter, too, right?

There's `TTree.lazyarray` (singular), `TTree.lazyarrays` (plural), and `TBranch.lazyarray` (singular). As you explained, the `uproot.lazyarray` and `uproot.lazyarrays` can't be unchunked (unless, of course, you `awkward.concatenate` them, but that's not obviously helpful. (Multi-file arrays might always or almost always be too big, unchunked.) And `uproot.daskarray` and `uproot.daskframe` shouldn't have this option because they _must_ be chunked.

Issue #376, opened 2019-10-14
Problems with identifying the type of a ROOT object


masonproffitt:
It seems unnecessarily difficult to figure out what an object's type is once read from a ROOT file. I often want to open a ROOT file and then scan across the TKeys to find the one(s) that corresponds to a TTree, for example. But this is what I run into with uproot:

```python
>>> import uproot
>>> file = uproot.open('run0.root')
>>> tree = file.values()[0]
>>> type(tree)
<class 'uproot.rootio.TTree'>
>>> isinstance(tree, uproot.rootio.TTree)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'uproot.rootio' has no attribute 'TTree'
```

I can do this:

```python
>>> isinstance(tree, uproot.tree.TTreeMethods)
True
```

but there's no intuitive reason to check TTreeMethods unless I dig into the documentation and eventually find that TTrees are a subclass of it.

It would be slightly better if I could do this:

```python
>>> tree._classname == 'TTree'
False
```

...but that also fails because I need to check against a byte literal:

```python
>>> tree._classname           
b'TTree'
>>> tree._classname == b'TTree'
True
```

which is also very non-intuitive.

Is there a way to make something like `isinstance(tree, uproot.TTree)` work, or to make `tree._classname == 'TTree'` work, or both? Why doesn't `uproot.open()` have a `namedecode` option like `arrays()`?

jpivarski:
It's also important to be able to identify the type of an object without reading it (i.e. to decide whether you want to read it). For that, there's `classes` as an alternative to `items`.

The Python class ones for each C++ class might already have a type name string (attached to the object itself), though I hadn't thought that it should be user-accessible. Maybe that should be exposed as a `classmethod` (what should the name be?).

The bytes-to-Unicode conversion and C++-to-Python name conversion are done at the same time—C++ names (with possible `[<>:]` characters) are raw bytes read from the ROOT file and Python names (pure `[a-zA-Z_][a-zA-Z_0-9]` and `str` type) are used as actual class names, converting `[<>:]` to distinct sequences (by losing `_` as an unencoded character). It sounds like you want Unicode names without the `[<>:]` translation.

Suggest an economic solution as a PR.

masonproffitt:
[...]

PR #379, opened 2019-10-16
Provide method to get ROOT class names as plain strings

masonproffitt:
See #376.


uproot/rootio.py
```diff
@@ -300,6 +304,9 @@ def items(self, recursive=False, filtername=nofilter, filterclass=nofilter):
     def classes(self, recursive=False, filtername=nofilter, filterclass=nofilter):
         return list(self.iterclasses(recursive=recursive, filtername=filtername, filterclass=filterclass))
 
+    def classnames(self):
+        return list(key._fClassName.decode('ascii') for key in self._keys)
```
jpivarski: This one should probably be a list of 2-tuples and take `recursive`, `filtername`, `filterclass` methods, like `classes`. That would make it easier to match up the output with the keys.

```python
    def iterclassnames(self, recursive=False, filtername=nofilter, filterclass=nofilter):
        for key in self._keys:
            if filtername(key._fName) and filterclass(cls):
                yield self._withcycle(key), key._fClassName
            if recursive and (key._fClassName == b"TDirectory" or key._fClassName == b"TDirectoryFile"):
                for name, classname in key.get().iterclassnames(recursive, filtername, filterclass):
                    yield "{0}/{1}".format(self._withoutcycle(key).decode("ascii"), name.decode("ascii")).encode("ascii"), classname

    def classnames(self, recursive=False, filtername=nofilter, filterclass=nofilter):
        return list(self.iterclassnames(recursive=recursive, filtername=filtername, filterclass=filterclass))

    def allclassnames(self, filtername=nofilter, filterclass=nofilter):
        return self.classnames(recursive=True, filtername=filtername, filterclass=filterclass)
```

uproot/rootio.py
```diff
@@ -1295,6 +1302,10 @@ def _recarray_dtype(cls, cntvers=False, tobject=True):
 
         return numpy.dtype(dtypesout)
 
+    @classmethod
+    def classname(cls):
+        return cls.__name__
```
jpivarski: I'm pretty sure this one is supposed to be

```python
    @property
    def classname(self):
        return self._classname._classname.decode('ascii')
```

Yes: [confirmed here](https://github.com/scikit-hep/uproot/blob/98b9758da71981ecccbe5c0ce040022331b4d7c2/uproot/rootio.py#L869-L872). If you return `cls.__name__`, then that would be the Python name of the class, which would disagree with the classnames defined above in cases with `<`, `>`, `:` characters. With these three `classname`/`classnames`, the user will get C++ names as `str` (with all `<`, `>`, `:` from templates, _not_ translated into underscores), which is what they expect.

masonproffitt: This is what I meant in the issue when I said I wasn't sure if it would work with complicated class names. I don't think I've ever encountered a `TKey` associated with a class with those characters in the name, but sure, it would be nice to get it to work in that case as well.

The problem here is having it as a `@property`. That means you can't access it from the class without an instance, which is something that I want to be able to do.

uproot/rootio.py
jpivarski: Maybe "classes" was the wrong name for this (because that implies just the classes, not also the key names), but we need the correlation because it's safer than trusting that we always get these keys in the same order. It would be nicer if it came back as a dict, but for many versions of Python, dicts don't preserve order, and I'd like to at least _try_ to preserve order.

Maybe that's not a great argument. I'm open to changing that interface in uproot 4, but for this PR, I'd like the interfaces to be consistent.

uproot/rootio.py
jpivarski: Right, that's why I dug into it, to at least check the code and try to do the right thing. I've encountered class names with `<`, `>`, `::` in bug reports, but I don't know if I have an example. (I'd like to just try to do the right thing now and if it's broken when we encounter an example, fix it then.)

It _is_ possible to access a property from a class object if that property does not depend on `self`:

```python
>>> class Something:
...     @property
...     def whatever(self):
...         return 999
... 
>>> Something().whatever
999
>>> Something.whatever.fget(None)
999
```

But this one does.

However, I don't see how you could ever get access to one of these classes without having an instance—they're generated from the streamers in the ROOT file: they're only available in `file._context.classes`, which I don't think you're digging into, right?

uproot/rootio.py
masonproffitt: Okay, I updated this. But is it not guaranteed that `.keys()` will return the same order as `.classes()`? I would have expected to do `for key, class in zip(f.keys(), f.classes()` if I wanted to match keys to classes (or class names).

uproot/rootio.py
masonproffitt: It's quite easy to access the class before you see the actual instance. The way #376 arose was that I was trying to do something like the following to get the class names of objects in a file:

```python
>>> import uproot
>>> file = uproot.open('055400.root')
>>> file.classes()[0][1]._classname
b'TTree'
>>> file.classes()[1][1]._classname
<property object at 0x7f7aeeb8fdd0>
```

As you can see, the TTree produces a `bytes` object and the second (a TNamed) returns a property, which is quite odd and unhelpful. It's true that I could instantiate the TNamed to get `_classname`, but that seems very clunky, especially since the behavior is different from the TTree and the fact that the class name is intrinsic to the class and not a true instance property.

jpivarski: Oh, darn, you're right. In that case, the way this should be done is to burn in another attribute on the class, at https://github.com/scikit-hep/uproot/blob/98b9758da71981ecccbe5c0ce040022331b4d7c2/uproot/rootio.py#L869-L872

```python
            if sys.version_info[0] > 2:
                code.insert(0, "    _classname = {0}".format(repr(streamerinfo._fName)))
            else:
                code.insert(0, "    _classname = b{0}".format(repr(streamerinfo._fName)))
            code.insert(0, "    classname = {0}".format(repr(streamerinfo._fName.decode("ascii")))
```

The `streamerinfo._fName` is the original C++ class name; we just need to decode that into a `str` and add it to the auto-generated code for the class.

masonproffitt: Right, that's what I started to do originally with this, but the problem is you would then have to deal with `builtin_classes` individually. That's because the code you pointed to only runs under this condition:

https://github.com/scikit-hep/uproot/blob/eb5d8862f8a382606be7bb9fa84f169b7cd19513/uproot/rootio.py#L676

For example, what you wrote works for TTree but not for TNamed. I'd have to go into the definition of each of these:

https://github.com/scikit-hep/uproot/blob/eb5d8862f8a382606be7bb9fa84f169b7cd19513/uproot/rootio.py#L1520-L1538

and add a `classname = ` line. This is why I instead tried to write a `classmethod`.

jpivarski: I decided to go with slightly WETter code than have this inconsistency. See 611308f81cb8e79f96e2b4359fa590600d414bf8 .

jpivarski:
If you want to talk about this before I merge it, find me today!

uproot/rootio.py
```diff
@@ -1452,24 +1499,38 @@ def tostring(self):
         return numpy.asarray(self, dtype=self._dtype).tostring()
 
 class TArrayC(TArray):
+    _classname = b"TArray"
+    classname = "TArray"
```
masonproffitt: Are these strings meant to be "TArrayC"?

jpivarski: Thanks! You're right, and this is the reason why _not_ to write WET code. However, trade-offs.

I did another sweep over all the `classname` assignments that exist and they all look right. The number of non-streamer class definitions won't be growing or won't be growing much. They are exceptions—most classes get defined from streamers.

Oh, I should add `classnames` (both the underscored and non-underscored version) for the classes that define the streamers themselves.

jpivarski:
Are you happy with me merging this?

masonproffitt:
Yeah, this looks fine to me.

PR #389, opened 2019-11-01
Laxer policy on branch conformance between files

### LINKED ISSUE #388: iterate does not read all requested branches
I'm reading multiple root files containing trees. Some of the root files contain branches, which the others don't. I only need selected branches from the root files. Some of the branches I want are only present in some root files. Thus I use uproot.iterate, where the branches argument is a function TBranchMethods ⇒ True or False.
What I would expect is, that the dict yielded by iterate contains the branches as they are available.
But instead the branches inside the dict are determined by the branch availabilities of the first file that was read: When the first file does not have some branches, they are not read for later files that have them; when a later file does not have a branch, an exception is raised by uproot/tree.py.
Everything seems to be caused by lines 178 to 183 in tree.py.
I can not understand this behaviour from the documentation. It also does not make sense to call the function given by the branches argument in every iteration step if the branches are determined by the first call. For my case, the only way I could solve the issue is by splitting the files I have into categories of their available branches and then having a iterate loop for each of the categories. This would be complicate things by a lot.

jrueb:
Fixes #388 
Also removed the old unused variables and renamed the "newbranches" variable, as there is no "oldbranches" variable anymore.


jrueb:
Sure, I think is fine to merge it. I'm glad I could help.

jpivarski:
Thanks!

PR #392, opened 2019-11-05
Set chunkbytes to 1 MB and limitbytes to 100 MB, based on @chrisburr's tests.

jpivarski:


jpivarski:
@lgray @nsmith- Could this PR be tested in Coffea to make sure it's not a terrible idea? Only after you're comfortable with it (with however much or little testing you dare) will I merge it.

Thanks!

jpivarski:
@nsmith- told me in person that it has a negligible impact (the concern here is that it would _worsen_ performance, so this is good news). Is that good enough for you, @lgray?

lgray:
@jpivarski Looking through recent probings with spark and the hadoop-xrootd connect, I was using larger values of both with no ill effects that I could find at scale (though slowdowns are coming from elsewhere).

nsmith-:
I revise a bit: the `uproot.numentries()` feature sees a noticable slowdown due to this, taking 3.3x the time (40s vs 12s with 4 worker process pool accessing FNAL EOS from an LPC login) to get entries for 1000 files.

nsmith-:
What I might do is edit the preprocessing step in coffea to set a more appropriate read chunk size.  The actual runtime doesn't seem to change much with the new read chunk size.

jpivarski:
> I revise a bit: the `uproot.numentries()` feature sees a noticable slowdown due to this, taking 3.3x the time (40s vs 12s with 4 worker process pool accessing FNAL EOS from an LPC login) to get entries for 1000 files.

This is good to know: I can set the default for `uproot.numentries` back to what it was before—we can have 1 MB on general reading and 32 kB on `uproot.numentries`. Would that do it?

nsmith-:
Well the real thing to do is actually record the chunk fetching behavior over some reasonable ensemble of files and tune it.

chrisburr:
The ideas #393 should be able to tune the request to only request the exact bytes which are needed so I don't think it's worth investing much time in finely tuning the chunksize itself.

jpivarski:
I guess it's "overtraining" that we're trying to avoid.

jpivarski:
I've put in the two policies. Is everybody happy with

   * `uproot.open` has `chunkbytes="1 MB", limitbytes="100 MB"`
   * `uproot.numentries` has `chunkbytes="32 kB", limitbytes="1 MB"`

@chrisburr, @nsmith-, @lgray?

Issue #393, opened 2019-11-05
HTTP multipart/XRootD vector_read for remote baskets


jpivarski:
The chunked sources (`HTTPSource` and `XRootDSource`) currently have a fixed-chunksize cache. Since the base directory and TTree metadata of most files is at the end of the file, this means that `uproot.open("http://...")` consists of two sequential remote reads: the beginning of the file (`TFile` header and all `TStreamerInfos`) and the end of the file, and the `file["tree"]` request probably hits the latter. This is a good argument for making the chunksize large (1 MB, PR #392), so that it is not likely to be more than two such reads.

However, when a user requests a single or a subset of branches from the TTree, a large chunksize might require too much data downloading because 1 MB is coarser than most basket sizes. Ideally, we want to send an HTTP multipart or XRootD vector_read of _only_ the bytes corresponding to the basket data. They are known before the read:

```python
branch._fBasketSeek[:branch.numbaskets]    # start of each basket in bytes
branch._fBasketBytes[:branch.numbaskets]   # size of each basket in bytes
```

The main chunked source cache relies on the fact that the chunksize is fixed, so that finding the appropriate chunk is a constant-time operation (it is performed many times, for every set of contiguous, fixed-size fields and for 2-3 times for every variable-size string). These baskets would be variable-sized, so they should be a _second cache_, in addition to the main cache. The second cache would have to maintain a list of start positions and sizes, with log-N lookup time via Python's `bisect` or Numpy's `searchsorted`.

Requesting data from baskets would have to be called in a different way for baskets than for every other case. This is not a major problem, since such access is already abstracted through `Cursor` and `Source`: a new method could be added to `Source`, like `Source.data` but going through the second cache (in chunked sources; no difference in other sources) and a new method could be added to `Cursor` for getting an array that is only used in basket reading.

The HTTP multipart or XRootD vector_read call could be added to `HTTPSource.preload` and `XRootDSource.preload` instead of the one-request-per-task parallel preloader that's in there now. The hook exists, but it will need to be expanded from knowing just the start of each seek point of interest to knowing the full ranges (starts and stops or starts and sizes).

[...]

PR #411, opened 2019-11-21
Reading files with HTTP fails if range headers are not supported

chrisburr:
Uproot fails to read files from servers that don't support range requests. Example traceback with extra debugging print out and without this fix applied:
```
Got response of size 20166303 chunksize is 1048576
-1765915941 -30694 -1671616656 4090602292 -7813 16221 -118717542 708660270
self._compressedbytes, self._uncompressedbytes = -1765908128 -1671616656
cursor.index - start 0
self._compressedbytes -1765908128
self._uncompressed = None
  File "<string>", line 1, in <module>
  File "uproot/rootio.py", line 60, in open
    return http(path, httpsource=httpsource, **options)
  File "uproot/rootio.py", line 89, in http
    return ROOTDirectory.read(openfcn(path), **options)
  File "uproot/rootio.py", line 154, in read
    streamerinfos, streamerinfosmap, streamerrules = _readstreamers(streamerkey._source, streamerkey._cursor, streamercontext, None)
  File "uproot/rootio.py", line 556, in _readstreamers
    tlist = TList.read(source, cursor, context, parent)
  File "uproot/rootio.py", line 949, in read
    out = cls._readinto(out, source, cursor, context, parent)
  File "uproot/rootio.py", line 1455, in _readinto
    start, cnt, self._classversion = _startcheck(source, cursor)
  File "uproot/rootio.py", line 426, in _startcheck
    cnt, vers = cursor.fields(source, _startcheck._format_cntvers)
  File "uproot/source/cursor.py", line 46, in fields
    return format.unpack(source.data(start, stop))
  File "uproot/source/compressed.py", line 183, in data
    return self._uncompressed[start:stop]
TypeError: 'NoneType' object is not subscriptable
```

Error can be seen when using Python's built in HTTP server to open a "large" file (i.e. more than one chunk):
```bash
python -m http.server 8000 --directory tests/samples
```

Also includes some minor fixes for linter errors I stumbled upon. It might be necessary to use a more general exception in some cases. I've not checked closely but perhaps `SyntaxError` needs to be caught in some cases?

### DIFF
--- a/uproot/interp/auto.py
+++ b/uproot/interp/auto.py
@@ -200,7 +200,7 @@ def interpret(branch, awkwardlib=None, swapbytes=True, cntvers=False, tobject=Tr
                             numbits = None
                         else:
                             low, high, numbits = spec
-                    except:
+                    except Exception:
                         return None
 
                 if iskDouble32 and numbits == 0:
--- a/uproot/interp/jagged.py
+++ b/uproot/interp/jagged.py
@@ -36,15 +36,12 @@ class asjagged(uproot.interp.interp.Interpretation):
         self.skipbytes = skipbytes
 
     def __repr__(self):
-        return "asjagged({0})".format(repr(self.content))
+        return "asjagged({0}{1})".format(repr(self.content), "" if self.skipbytes == 0 else ", {0}".format(self.skipbytes))
 
     def to(self, todtype=None, todims=None, skipbytes=None):
         if skipbytes is None:
             skipbytes = self.skipbytes
         return asjagged(self.content.to(todtype, todims), skipbytes)
-
-    def __repr__(self):
-        return "asjagged({0}{1})".format(repr(self.content), "" if self.skipbytes == 0 else ", {0}".format(self.skipbytes))
 
     @property
     def identifier(self):
--- a/uproot/source/chunked.py
+++ b/uproot/source/chunked.py
@@ -123,7 +123,9 @@ class ChunkedSource(uproot.source.source.Source):
 
             if len(chunk) > self._chunkbytes:
                 if len(chunk) > self._limitbytes:
-                    warnings.warn('Received larger chunk than expected, assumed to be entire file. Performance may be significantly degraded.', RuntimeWarning)
+                    warnings.warn('Remote server sent a larger chunk than expected, assumed to be entire file. '
+                                  'Performance will be degraded as the file is larger than the cache limit and might need to be downloaded many times.',
+                                  RuntimeWarning)
                 for i in range(0, len(chunk), self._chunkbytes):
                     if i // self._chunkbytes not in self.cache:
                         self.cache[i // self._chunkbytes] = chunk[i:i+self._chunkbytes]
--- a/uproot/source/chunked.py
+++ b/uproot/source/chunked.py
@@ -125,6 +125,8 @@ class ChunkedSource(uproot.source.source.Source):
                 warnings.warn('Received larger chunk than expected, assumed to be entire file. Performance may be significantly degraded.', RuntimeWarning)
                 for i in range(0, len(chunk), self._chunkbytes):
                     self.cache[i//self._chunkbytes] = chunk[i:i+self._chunkbytes]
+                # Dismiss any pending futures as everything has already been loaded
+                self.dismiss()
                 chunk = self.cache[chunkindex]
             else:
                 self.cache[chunkindex] = chunk
--- a/uproot/source/chunked.py
+++ b/uproot/source/chunked.py
@@ -5,7 +5,6 @@
 from __future__ import absolute_import
 
 import math
-import warnings
 
 import numpy
 
@@ -122,18 +121,12 @@ class ChunkedSource(uproot.source.source.Source):
             gstop = (chunkindex + 1) * self._chunkbytes
 
             if len(chunk) > self._chunkbytes:
-                if len(chunk) > self._limitbytes:
-                    warnings.warn('Remote server sent a larger chunk than expected, assumed to be entire file. '
-                                  'Performance will be degraded as the file is larger than the cache limit and might need to be downloaded many times.',
-                                  RuntimeWarning)
+                if not numpy.array_equal(chunk[:4], list(b"root")):
+                    raise NotImplementedError("Expected {0} or fewer bytes but received {1} and data does not appear to be an entire ROOT file.".format(self._chunkbytes, len(chunk)))
+                self.cache = {}
                 for i in range(0, len(chunk), self._chunkbytes):
-                    if i // self._chunkbytes not in self.cache:
-                        self.cache[i // self._chunkbytes] = chunk[i:i+self._chunkbytes]
-                # Add the requested range last to minimise cache misses
-                for i in range(chunkstart*self._chunkbytes, chunkstop*self._chunkbytes):
-                    if i // self._chunkbytes not in self.cache:
-                        self.cache[i // self._chunkbytes] = chunk[i:i+self._chunkbytes]
-                chunk = chunk[gstart:gstop]
+                    self.cache[i // self._chunkbytes] = chunk[i:i+self._chunkbytes]
+                chunk = self.cache[chunkindex]
                 # Dismiss any pending futures as everything has already been loaded
                 self.dismiss()
             else:
--- a/uproot/source/chunked.py
+++ b/uproot/source/chunked.py
@@ -5,11 +5,13 @@
 from __future__ import absolute_import
 
 import math
+import warnings
 
 import numpy
 
 import uproot.cache
 import uproot.source.source
+
 
 class ChunkedSource(uproot.source.source.Source):
     # makes __doc__ attribute mutable before Python 3.3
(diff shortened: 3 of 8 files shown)


uproot/source/chunked.py
```diff
@@ -106,21 +108,29 @@ def data(self, start, stop, dtype=None):
...
-                    chunk = self.cache[chunkindex] = self._read(chunkindex)
+                    chunk = self._read(chunkindex)
 
             cstart = 0
             cstop = self._chunkbytes
             gstart = chunkindex * self._chunkbytes
             gstop = (chunkindex + 1) * self._chunkbytes
 
+            if len(chunk) > self._chunkbytes:
+                warnings.warn('Received larger chunk than expected, assumed to be entire file. Performance may be significantly degraded.', RuntimeWarning)
```
jpivarski: I'd rather not have warnings, even if performance is significantly degraded. (There's an uncountable number of ways to do _that_, after all.) Exceptions in Python are fine—they tell you where to go, or at least start looking, to fix them—but warnings are "sticky"—you see them but don't know how to fix them (if you don't search the codebase for the text of the warning).

uproot/tree.py
```diff
@@ -1816,17 +1816,17 @@ def _init(self):
...
+                "treepath": self.treepath,
+                "branches": self.branches,
+                "entrysteps": self.entrysteps,
+                "flatten": self.flatten,
+                "awkwardlib": self.awkwardlib,
+                "persistvirtual": self.persistvirtual,
+                "localsource": self.localsource,
+                "xrootdsource": self.xrootdsource,
+                "httpsource": self.httpsource,
+                "options": self.options}
```
jpivarski: I wonder how this happened... someone must have asked for `_LazyFiles` to be pickleable, and I added this half-hearted attempt without checking it? Thanks for fixing it.

uproot/source/compressed.py
```diff
@@ -119,7 +119,7 @@ def _prepare(self):
                 # https://github.com/root-project/root/blob/master/core/zip/src/RZip.cxx#L217
                 # https://github.com/root-project/root/blob/master/core/lzma/src/ZipLZMA.c#L81
                 # https://github.com/root-project/root/blob/master/core/lz4/src/ZipLZ4.cxx#L38
-                algo, method, c1, c2, c3, u1, u2, u3 = cursor.fields(self._compressed, self._header)
+                algo, method, c1, c2, c3, u1, u2, u3 = header = cursor.fields(self._compressed, self._header)
```
jpivarski: I see that `header` gets used on line 151 below.

uproot/source/chunked.py
```diff
@@ -5,12 +5,14 @@
 from __future__ import absolute_import
 
 import math
+import warnings
```
jpivarski: See rant about warnings below.  :)

uproot/interp/jagged.py
```diff
@@ -36,16 +36,13 @@ def __init__(self, content, skipbytes=0):
...
+        return "asjagged({0}{1})".format(repr(self.content), "" if self.skipbytes == 0 else ", {0}".format(self.skipbytes))
 
     def to(self, todtype=None, todims=None, skipbytes=None):
         if skipbytes is None:
             skipbytes = self.skipbytes
         return asjagged(self.content.to(todtype, todims), skipbytes)
 
-    def __repr__(self):
-        return "asjagged({0}{1})".format(repr(self.content), "" if self.skipbytes == 0 else ", {0}".format(self.skipbytes))
-
```
jpivarski: Oops; thanks!

uproot/interp/auto.py
```diff
@@ -200,7 +200,7 @@ def transform(node, tofloat=True):
                             numbits = None
                         else:
                             low, high, numbits = spec
-                    except:
+                    except Exception:
```
jpivarski: We can be explicit about the classes caught by general exception handlers. (Is that in a style guide? I'll start doing it, if so.)

chrisburr: I'm going to quote from PEP8 rather than writing the same thing less clearly:

> When catching exceptions, mention specific exceptions whenever possible instead of using a bare `except:` clause:
> ```python
> try:
>     import platform_specific_module
> except ImportError:
>    platform_specific_module = None
> ```
> A bare `except:` clause will catch `SystemExit` and `KeyboardInterrupt` exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use `except Exception:` (bare except is equivalent to `except BaseException:`).
> 
> A good rule of thumb is to limit use of bare 'except' clauses to two cases:
> 
>    * If the exception handler will be printing out or logging the traceback; at least the user will be aware that an error has occurred.
>    * If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise. try...finally can be a better way to handle this case.

uproot/source/chunked.py
chrisburr: The issue with performance here when the cache size limit is smaller than the entire file. My fix is very naive and will just keep downloading the same file over and over and over again while uproot seeks through it.

Now I've thought about it again, I've just pushed an improvement to make the situation slightly less bad. I also now ensure that even if the current chunk is evicted from the cache it will still work.

I see three options for how to handle this case:
1. Only show the warning if the file is larger than the cache? To me it feels like enough of a footgun that users should know but you're right that it isn't clear what to do. I've just done this and edited the message to be more explicit.
2. Alternatively as the entire file is already in memory at this point, the cache could be replaced with a simple dictionary containing the entire file to waste memory instead of time?
3. Drop the warning entirely like you first suggested.

Which would you prefer?

jpivarski:
I think the reason performance is usually treated as secondary it's because it's so dependent on the use-case. What you're calling a "performance degradation" here is based on assumption that the user will want to read nearby arrays from the file. (And with that basket-sized cache blocks improvement we talked about, "the same array more than once.") This is a common use-case, though not the only one, and a warning about performance for those users would be very confusing.

What I had in mind for remote files was to reproduce the memory map API that works so well on local files. If the file is too big for memory—a common situation in analysis, though not tutorials—the cache will thrash. This is the case with all kinds of caches, and it's the reason that single pass (e.g. doing all your work in an `uproot.iterate` loop) is faster.

Is the situation here that the cache loaded to read basket TKeys will get evicted before we can read the array data? It's one thing if the cache thrashes when the user asks for the same data twice, but quite another if it thrashes while servicing a single request.

jpivarski:
Oh! What's happening is that you ask HTTP for a chunk, but it answers by sending the whole file? If we know that, why don't we chop it up into the right-sized chunks send fill the cache? In other words, take the error as a gift?

Why would HTTP do that? Does it only happen for small files? I imagine it would be a problem if asking for a byte range results in downloading TB of unwanted data.

To do this, we'd have to be certain of where the downloaded part starts. If `len(chunk) > chunksize`, does that mean that it starts at `0` or at the requested `start`?

uproot/source/chunked.py
jpivarski: See comments below.

chrisburr:
I should have been clearer as this is not a solution for #393, it's just a bug I noticed when starting to work on it.

> Oh! What's happening is that you ask HTTP for a chunk, but it answers by sending the whole file?

Yes! Sorry I really should have said that in the first comment of this PR. I'll blame jet lag as I've just got back from CHEP 😄 

> Why would HTTP do that? Does it only happen for small files?

It's dependent on if the server. I've not looked into it enough to know exactly but as range requests are just a header in the request, if the HTTP server is very simple like `python -m http.server` (or if something is misconfigured and the header gets lost) the header is ignored and you get the entire file.

> If we know that, why don't we chop it up into the right-sized chunks send fill the cache? In other words, take the error as a gift?

That is what is currently done in this PR. The big performance issue if the cache is smaller than the file. The cache becomes useless when trying to read almost anything in a 4GB file with the default cache size of 100MB and you'll download 4GB for every single call to `Source.data`. The newest two commits make it a little better if you only read nearby data (this isn't the case in my test files).

> I imagine it would be a problem if asking for a byte range results in downloading TB of unwanted data.

This is why I suggested a warning in option 1 as it's not just "bad performance for some usage patterns" it's repeatedly reading millions of times more bytes than you need to and not being able to cache them effectively unless you only access the same chunk or one at the end of the file.

The alternative I see is to ignore the cache size when this happens and just use a normal dictionary (option 2). It's not nice though as memory usage might blow up. Personally I find that more acceptable to do without a warning as it's easier to debug when it causes a problem.

> To do this, we'd have to be certain of where the downloaded part starts. If len(chunk) > chunksize, does that mean that it starts at 0 or at the requested start?

Empirically this seems to start at zero as if the header was ignored and I don't see how there could be any other reasonable behaviour. Not being sure about this was another reason to include the warning. Are there any magic bytes at the start of a ROOT file which I can look for to check this assumption?

jpivarski:
> Are there any magic bytes at the start of a ROOT file which I can look for to check this assumption?

Yes!

```bash
% hexdump -C issue412.root | head -1
00000000  72 6f 6f 74 00 10 33 a8  00 00 00 64 00 00 00 00  |root..3....d....|
```

The first four characters of every ROOT file are guaranteed to be `"root"` (that aspect of the format _is_ [specified](https://root.cern.ch/doc/master/classTFile.html)). If character data in a ROOT file were uniformly random, the chance of hitting these four characters by accident is about 1 in 4 billion. It's not uniformly random; there are a few places in the file where the filename is stored, which probably ends in `".root"`. (Surprise! If you name your file a dirty word while filling it, then change its name on the commandline, it will have that dirty word embedded inside the file. Another fun quirk.) If HTTP returns a _random_ fragment of the file (which I'd consider a bug), then there is a small chance that those non-beginning bytes are `"root"`. Not as small as 1 in 4 billion, but negligible.

The `"root"` magic is followed by a 4-byte big-endian version number (such as `61800` for 6.18/00, possibly with `1000000` added if it's a big file, like `1061800`). The next 4 bytes (if small) or 8 bytes (if big) is `fBEGIN`, which is almost always `100`. Maybe these could be used to disambiguate—it's not likely to be randomly after `"root"`, but perhaps that's too complex to check for.

Your suggestion of increasing the cache size is dangerous because people _do_ run out of memory, and these cases are hard to track down. Making the cache behavior more complex will make those cases more difficult to debug. However, if we're in this situation where we've unexpectedly downloaded the whole file and Python is still alive, then we've demonstrably fit the whole thing into memory. Casting it as an array with `numpy.frombuffer` and slicing it into tiles is a zero-copy operation that wouldn't add to the total usage.

How about this: if HTTP returns a whole file instead of the requested part, which we can determine by `len(chunk) > chunksize and chunk[:4] == b"root"`, then let's chop it up into the correct `chunksize` (assuming that it starts at `0`) and put all of the chunks into a simple-dictionary cache, rather than the eviction-happy cache. After all, evicting a view doesn't even reduce memory usage, so the normal rules don't apply here anyway.

So, yes, that's option 2. I'm on board now.

chrisburr:
Thank you for the interesting write up. (As always!)

Checking for `b"root"` should be adequate as checking for a sensible version seems either overly general or fragile. If the check doesn't pass it raises a `NotImplementedError`.

The response from `Source._read` is already put through `return numpy.frombuffer(data, dtype=numpy.uint8)` so I think this is now ready.

chrisburr:
@jpivarski Assuming you're happy with it, this can be merged.

As nobody has reported it as a bug I think it can wait for the next release instead of having a dedicated one made.

jpivarski:
Okay, thanks!

PR #417, opened 2019-12-10
Embed branch titles in awkward docstrings for lazyarrays

nsmith-:
Although python has nothing to say about instance docstrings, it seems IPython supports setting `__doc__` on both class and class instances, in which case it displays both with the `?` magic or shift+tab in jupyter.  This PR takes advantage of that to display branch titles in lazy array docstrings.
This might not work in python 2, not sure.  If not, I'll just use `six` to disable this feature for py2.

### DIFF
--- a/uproot/tree.py
+++ b/uproot/tree.py
@@ -1621,7 +1621,7 @@ class TBranchMethods(object):
             return chunkedarray(chunks, counts)
         else:
             start, stop = entrysteps[0]
-            return VirtualArray(lazybranch, (start, stop), cache=cache, type=awkward.type.ArrayType(numentries, interpretation.type), persistvirtual=persistvirtual)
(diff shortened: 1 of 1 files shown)


jpivarski:
Chances are, it just doesn't do anything in Python 2. Old Python docstrings were more "hard-coded," and attempts to set them with `__doc__` (at the class level) were just ignored. There was a hack to fix this by giving the class a custom metatype, which you can see in uproot.

Uproot doesn't depend on `six`, so if we had to set a switch, I'd rather just check `sys.versioninfo` (as you see elsewhere).

Is the PR done? I think `lazyarrays` gets checked in the tests, which passed (including Python 2).

nsmith-:
The PR is done, feature-wise.  It seems to work in my tests.

PR #421, opened 2019-12-16
Add zstd compression read support for uproot

oshadura:
Locally I have green tests:
Ran 106 tests in 33.114s
OK


uproot/source/compressed.py
```diff
@@ -80,6 +82,13 @@ def decompress(self, source, cursor, compressedbytes, uncompressedbytes=None):
                 raise ValueError("lz4 needs to know the uncompressed number of bytes")
             return lz4_decompress(cursor.bytes(source, compressedbytes), uncompressed_size=uncompressedbytes)
 
+        elif self.algo == uproot.const.kZSTD:
+            try:
+                from zstd import decompress as zstd_decompress
+            except ImportError:
+                raise ImportError("Install zstd package with:\n    pip install zstd\nor\n    conda install -c anaconda zstd")
```
jpivarski: The zstd package is in the `conda-forge` channel, not `anaconda`, and I just tried installing it: it appears to be the binary library (libstd.so), not the Python bindings. I think users will be forced to install the Python bindings through pip.

@chrisburr, do you happen to know the status of the zstd Python bindings in conda-forge? Is it coming, does it exist under a different name, or is it just not available?

appveyor.yml
```diff
@@ -50,7 +50,7 @@ build_script:
   - "python -c \"import awkward; print(awkward.__version__)\""
   - "pip install \"uproot-methods>=0.7.0\""
   - "python -c \"import uproot_methods; print(uproot_methods.__version__)\""
-  - "pip install cachetools lz4 mock xxhash"
+  - "pip install cachetools zstd mock xxhash"
```
chrisburr: I guess this is a typo:
```suggestion
  - "pip install cachetools lz4 zstd mock xxhash"
```

jpivarski: zstd shouldn't replace lz4; both should be added here.

jpivarski: You were faster than me! I included this in my review below. (Resolving for duplication.)

Do you know about availability of the zstd Python bindings in conda? (The other comment down below.)

uproot/source/compressed.py
chrisburr: Looks like there missing but it should be trivial to add. I'll do it later today as `python-zstd`.

I would also suggest editing the `lz4` message above to suggest `conda-forge` instead of `anaconda` (or not suggest a channel at all).

jpivarski: I think that's a good idea: removing the `-c conda-forge` from the lz4 and lzma messages (although the lzma message is only triggered in Python 2.7, so that's becoming legacy). If the user uses conda to install packages, they must have already added conda-forge as a channel, otherwise they would not have gotten uproot. The [uproot front-page](https://github.com/scikit-hep/uproot#installation) recommends adding conda-forge as a channel.

oshadura: > The zstd package is in the `conda-forge` channel, not `anaconda`, and I just tried installing it: it appears to be the binary library (libstd.so), not the Python bindings. I think users will be forced to install the Python bindings through pip.
> 
> @chrisburr, do you happen to know the status of the zstd Python bindings in conda-forge? Is it coming, does it exist under a different name, or is it just not available?

@chrisburr  @jpivarski thanks a lot for an input! I will fix PR accordingly.

appveyor.yml
oshadura: @jpivarski sorry, it was definitely a typo!

oshadura:
@jpivarski I will add zstd for writing as soon we will have release.

jpivarski:
@oshadura Writing with zstd should probably be a new pull request, and it may be after the 3.11.0 release.

When you do add writing with zstd, be sure to ping @reikdas so he can verify that nothing was missed.

jpivarski:
@oshadura I pushed a correction to your branch, but it hasn't shown up here yet (for me): https://github.com/oshadura/uproot/commits/add-zstd-compression-support

When it does (and tests pass), I'll merge this unless you tell me not to.

jpivarski:
@oshadura Could you try pulling in your branch, verifying that my changes to uproot/source/compressed.py are there (where I dropped the `-c XYZ` option in the install messages), and maybe make some trivial change to push it here? I don't know why it isn't showing up.

chrisburr:
I was looking at adding the conda-forge package and I noticed there are two different sets of Python bindings. Both are linked from [facebook's webpage](https://facebook.github.io/zstd/):
* https://pypi.org/project/zstd/: Currently used
* https://pypi.org/project/zstandard/: More downloads/GitHub activity and already on conda-forge

Is there a reason for preferring `zstd`?

oshadura:
@jpivarski 

> I was looking at adding the conda-forge package and I noticed there are two different sets of Python bindings. Both are linked from [facebook's webpage](https://facebook.github.io/zstd/):
> 
>     * https://pypi.org/project/zstd/: Currently used
> 
>     * https://pypi.org/project/zstandard/: More downloads/GitHub activity and already on conda-forge
> 
> 
> Is there a reason for preferring `zstd`?

Thanks, let me check!

oshadura:
> https://pypi.org/project/zstd/

@jpivarski @chrisburr thanks for checking, zstandard package is definitely better! (it supports much more functionality, such as dictionary generation and streaming API). I updated PR accordingly.

jpivarski:
Okay, great! When the tests pass, should I merge this, @oshadura?

oshadura:
> Okay, great! When the tests pass, should I merge this, @oshadura?

Yes, please!

jpivarski:
After this last force-push?

> oshadura force-pushed the oshadura:add-zstd-compression-support branch from 1d8af06 to 83169d9 2 minutes ago

oshadura:
@jpivarski yeap, now I am ready I think.

PR #427, opened 2019-12-19
Allow compressing ROOT objects which have size greater than 2^24

### LINKED ISSUE #333: Unable to compress data larger than 2^24 bytes
Data corresponds to individual objects like TTree, TH and each basket of a TTree.

reikdas:
Fixes #333

### DIFF
--- a/tests/test_write.py
+++ b/tests/test_write.py
@@ -1793,3 +1793,25 @@ def test_tree_cycle(tmp_path):
         assert branch1[i] == a[i]
         assert branch2[i] == b[i]
 
+def test_large_compress(tmp_path):
+    filename = join(str(tmp_path), "example.root")
+
+    with uproot.recreate(filename, uproot.ZLIB(5)) as f:
+        f["a"] = "a" * ((2 ** 24) + 2000)
+        f["b"] = "b" * ((2 ** 24) + 10)
+
+    f = ROOT.TFile.Open(filename)
+    assert str(f.Get("a")) == "a" * ((2 ** 24) + 2000)
+    assert str(f.Get("b")) == "b" * ((2 ** 24) + 10)
+    f.Close()
+
+def test_large_compress_uproot(tmp_path):
+    filename = join(str(tmp_path), "example.root")
+
+    with uproot.recreate(filename, uproot.ZLIB(5)) as f:
+        f["a"] = "a"*((2**24) + 2000)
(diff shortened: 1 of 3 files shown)


reikdas:
@jpivarski I think the tests are failing on travis because travis is unable to allocate ((2 ** 24) + 2000) bytes of contiguous memory. What do you think we should do?

jpivarski:
If it has been locally tested and only intermittently fails on Travis for the reason you suggested, I think we can remove this test, or at least give it a tag so that it can be selected with pytest, but not run automatically.

I restarted the failed job to see if it is intermittent.

jpivarski:
Yes. It was intermittent. I recommend tagging it as "bigtest" or something. There are other tests that could use a tag like that; this will be the first one.

reikdas:
> I restarted the failed job to see if it is intermittent.

The tests passed when you restarted the job :) I think this is good to go now.

reikdas:
When you say tag, do you mean this - https://github.com/scikit-hep/uproot/pull/427/files#diff-5f81591ed2ee25a817de2554fd35f608R1796 ? If that is the case, I think this PR is done.

reikdas:
Oh I just noticed something! Skipped tests act like they are an error because of this - https://github.com/scikit-hep/uproot/blob/master/.travis.yml#L72, so pytest.mark.skip does not work..

jpivarski:
I mean something like this: https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

Except that the large memory tests (or, if we ever add them, long-running tests) are only executed if desired: opt-in, rather than opt-out.

chrisburr:
> I think the tests are failing on travis because travis is unable to allocate ((2 ** 24) + 2000) bytes of contiguous memory. What do you think we should do?

16MB doesn't seem unreasonable and I'm [not able to reproduce that issue](https://travis-ci.org/chrisburr/uproot/builds/627560782) when reverting 
b03a52e. Do you have an example of it failing? I found the previous build but there has been force pushes since then.

reikdas:
It seems like it was a travis anamoly, I have reverted the commit and the tests seem to be working :)

jpivarski:
If everything looks good now, would you like me to merge it?

reikdas:
Hi, yes seems good.

PR #428, opened 2019-12-20
 Get pytest-error-for-skips from PyPI 

chrisburr:
I wasn't expecting my PR to fix `pytest-error-for-skips` to be accepted as it hadn't been updated in years but it has been so we can now get it from PyPI.

Also skips `test_issue390` if `pandas` is unavailable.


chrisburr:
Yeah, I'm very happy the maintainer jumped in to make a new release as it seems like a useful little package. I'd normally say it was too small to justify a dependency but as it's only for the CI I think it's justified.

I'm done so go ahead and merge!

PR #435, opened 2020-01-11
Allow dtypes with byteorders via lists. Fixes #434

### LINKED ISSUE #434: dtype byteswap
I have some trouble with the dtype in the `KM3NeT_TIMESLICE` format (see for example #433). The data can be read using 
```python
f = uproot.open("file.root")
tree = f[b'KM3NET_TIMESLICE_L2'][b'KM3NETDAQ::JDAQTimeslice']
superframes = tree[b'vector<KM3NETDAQ::JDAQSuperFrame>']
hits_buffer = superframes[
b'vector<KM3NETDAQ::JDAQSuperFrame>.buffer'].lazyarray(
    uproot.asjagged(uproot.astable(
        uproot.asdtype([("pmt", "u1"), ("tdc", "u4"),
              ("tot", "u1")])),
           skipbytes=6),
        basketcache=uproot.cache.ThreadSafeArrayCache(23*1024**2))
```
(taken from #433 ... thanks @tamasgal ;).

I tried to modify this code and use the dtypes "<u4" and ">u4" for the "tdc" field, in order to invert byteorder. Unfortunately this modification did not have any effect.

[example.zip](https://github.com/scikit-hep/uproot/files/4046711/example.zip)

tamasgal:
This patch fixes the byte order interpretation of dtypes which are passed as `list`s. The current implementation ignores them and falls back to big endian in case of the ROOT data (`fromdtype`) and native for the Python side (`todtype`).

Note that the workaround is to pass `numpy.dtype` instances, which works fine, so this patch is more like cosmetics, but can potentially prevent some confusion in future.


tamasgal:
Great, yes please merge then 😉

PR #437, opened 2020-01-15
Clear ChunkedSource cache on explicit close

nsmith-:
For those that want to ensure on-closure (or context managers)
cleanup of the (100 MB large by default) chunk cache as early as possible,
since the source is attached to uproot objects with reference cycles and won't be cleaned
up until the next generational gc run in python, which can lead to
unnecessarily high peak memory usage.

### DIFF
--- a/uproot/source/chunked.py
+++ b/uproot/source/chunked.py
@@ -46,6 +46,10 @@ class ChunkedSource(uproot.source.source.Source):
     def _read(self, chunkindex):
         raise NotImplementedError
 
+    def close(self):
+        super(ChunkedSource, self).close()
+        self.cache.clear()
+
     def dismiss(self):
         if self._futures is not None:
             for future in self._futures.values():


nsmith-:
I am not sure, but I might have requested it as I use it in https://github.com/CoffeaTeam/coffea/blob/master/coffea/processor/executor.py#L461
I know we discussed putting this call in `ROOTDirectory.__exit__` but since it might cause trouble for mmap sources we didn't make it a blanket thing.

jpivarski:
Maybe the thing to do is to make sure that `MemmapSource` doesn't actually close anything when its `close` is called:

https://github.com/scikit-hep/uproot/blob/2ac123fa14df312c14f7807eaa5219f099f26e7c/uproot/source/memmap.py#L35-L36

and have the `with` syntax actually call `close`. Currently, the `with` syntax does nothing (below is in `ROOTDirectory`):

https://github.com/scikit-hep/uproot/blob/2ac123fa14df312c14f7807eaa5219f099f26e7c/uproot/rootio.py#L373-L377

PR #441, opened 2020-01-24
Add PyPI and Conda-Forge badges to README

### LINKED ISSUE #440: Add PyPI and Conda-forge badges
I was recently asked for a quick way to find the latest version of Uproot, and I tried to point the user at the PyPI badge on the repo page, but there wasn't one. PyPI and Conda-forge version badges would be a useful addition!

matthewfeickert:
Resolves #440 

Add the following badges to the README:

- [PyPI version](https://pypi.org/project/uproot/)
- [Conda-Forge](https://github.com/conda-forge/uproot-feedstock)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,6 +17,15 @@ uproot
 .. image:: https://mybinder.org/badge_logo.svg
    :target: https://mybinder.org/v2/gh/scikit-hep/uproot/master?urlpath=lab/tree/binder%2Ftutorial.ipynb
 
+|PyPI version|
+|Conda-Forge|
+
+.. |PyPI version| image:: https://badge.fury.io/py/uproot.svg
+   :target: https://badge.fury.io/py/uproot
+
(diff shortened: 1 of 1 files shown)


henryiii:
> lose the feeling that they're useful

They provide several things. First is a way to quickly check the version - GitHub does not provide a way to see what the latest version is from the main page unless you add a GitHub releases badge - for a Python package, nobody really cares what the GitHub latest version is, though, they care about pip or conda. It also shows that the conda-forge and pypi versions are in sync (or when they are not). I use it to quickly check when I'm setting a minimum requirements.txt version (PyPI) or environments.yml (conda).

It also is useful to be able to get to the PyPI / Conda-forge pages sometimes - it's the only way to get to the conda-forge page for someone wanting to look at the recipe. I really hate packages that do not link to all the places that have related code, as it made it much harder when I was learning how things like conda-forge worked. And the PyPI page has a little more metadata that might be useful to someone. Some packages use the GitHub README for more dev-focused info, and the PyPI page for user-focused intros.

PR #442, opened 2020-01-27
Allow tests to be ran with minimal dependencies

chrisburr:
Adds more `importorskip` statements so the test suite can be ran without any of the optional dependencies (useful for the `uproot-base` package in conda-forge https://github.com/conda-forge/uproot-feedstock/pull/58).

Interestingly I see a failure with Python 3.6, 3.7 and 3.8 (but not 2.7!) on Windows which looks like it might be real ([full log](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=114956&view=logs&j=5be07ae1-d8ba-5406-47b6-8e3a3a12f825&t=0d1d4714-3221-52cd-cfdb-f62a891e036b)). Any guesses?

```
================================== FAILURES ===================================
______________________________ Test.test_issue38 ______________________________

self = <tests.test_issues.Test object at 0x000001CEBC425748>

    def test_issue38(self):
        before_hadd = uproot.open(
            "tests/samples/issue38a.root")["ntupler/tree"]
        after_hadd = uproot.open("tests/samples/issue38b.root")["ntupler/tree"]
    
        before = before_hadd.arrays()
        after = after_hadd.arrays()
    
        assert set(before.keys())
        assert set(after.keys())
    
        for key in before.keys():
>           assert before[key].tolist() * 3 == after[key].tolist()
E           assert [[0, 1, 0, 2,..., 0, 2, 0, 3]] == [[1, 2, 3], [...3], [1, 2, 3]]
E             At index 0 diff: [0, 1, 0, 2, 0, 3] != [1, 2, 3]
E             Use -v to get the full diff

tests\test_issues.py:80: AssertionError
================== 1 failed, 78 passed, 16 skipped in 10.78s ==================
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -44,6 +44,7 @@ install:
   - "python --version"
 
 build_script:
+  - "SET PATH=%PYTHON%;%PATH%"
   - "python -m pip install --upgrade pip"
   - "pip install %NUMPY%"
   - "pip install \"awkward>=0.12.0,<1.0\""
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,18 +41,18 @@ environment:
       NUMPY: "numpy>=1.15"
 
 install:
+  - "SET PATH=%PYTHON%;%PATH%"
   - "python --version"
 
 build_script:
-  - "SET PATH=%PYTHON%;%PATH%"
   - "python -m pip install --upgrade pip"
-  - "pip install %NUMPY%"
-  - "pip install \"awkward>=0.12.0,<1.0\""
+  - "python -m pip install %NUMPY%"
+  - "python -m pip install \"awkward>=0.12.0,<1.0\""
   - "python -c \"import awkward; print(awkward.__version__)\""
-  - "pip install \"uproot-methods>=0.7.0\""
+  - "python -m pip install \"uproot-methods>=0.7.0\""
   - "python -c \"import uproot_methods; print(uproot_methods.__version__)\""
-  - "pip install cachetools lz4 zstandard mock xxhash"
-  - "pip install -i https://pypi.anaconda.org/carlkl/simple backports.lzma"
-  - "pip install pytest pytest-runner pandas requests"
+  - "python -m pip install cachetools lz4 zstandard mock xxhash"
+  - "python -m pip install -i https://pypi.anaconda.org/carlkl/simple backports.lzma"
+  - "python -m pip install pytest pytest-runner pandas requests"
   - "curl -fsS -o tests/samples/Event.root http://scikit-hep.org/uproot/examples/Event.root"
   - "pytest -v tests"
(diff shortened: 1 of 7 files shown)


jpivarski:
The `importorskip` is a good idea. Just digging into it with

jpivarski:
Just digging into the issue with my phone, I can see that the error is here:

https://github.com/scikit-hep/uproot/blob/master/tests/test_issues.py#L68-L80

and it looks like an issue where NumPy in different platforms has different interpretations of dtypes as 32 or 64-bit. This issue has come up in other places, and it's often tied to Python 2 vs 3. That's why the test fails with

`[0, 1, 0, 2, 0, 3] != [1, 2, 3]`

jpivarski:
Is there any way we could log into one of these machines? I just reproduced the correct result (on Linux) and I see that this file has a suite of branch types. It's one of these:

```
v_int16                    TStreamerSTL               asjagged(asdtype('>i2'), 10)
v_int32                    TStreamerSTL               asjagged(asdtype('>i4'), 10)
v_int64                    TStreamerSTL               asjagged(asdtype('>i8'), 10)
v_uint16                   TStreamerSTL               asjagged(asdtype('>u2'), 10)
v_uint32                   TStreamerSTL               asjagged(asdtype('>u4'), 10)
v_uint64                   TStreamerSTL               asjagged(asdtype('>u8'), 10)
```

but probably a 64-bit one because of Windows-NumPy confusions about 32 and 64-bit. The testing matrix varies NumPy versions; does this only affect certain versions?

chrisburr:
I reproduced it in a VM and it is the `v_uint64` key that fails. I then noticed that Appveyor is using Python 2 for all of the tests I'll see if I can fix the CI after lunch.

chrisburr:
Okay now I've "fixed" the Appveyor CI I can see the same thing there. If it's helpful for debugging I can try things in my Windows VM but I don't have time at the moment to properly dig into the cause.

jpivarski:
After spending all morning installing Windows and Anaconda on a VM (talking with Cortado), I was able to work interactively on the problem and found it in a couple of minutes.  `:)`  I'm going to keep this VM around.

The issue was that we were interpreting the `kLong` and `kULong` types (not `kLong64` and `kULong64`) in a platform-dependent way:

https://github.com/scikit-hep/uproot/blob/ced6380f7dcf558f7d7f20a5673514479bd37efa/uproot/interp/auto.py#L54-L60

Note that `numpy.long` is equal to `numpy.int32` on 32-bit Windows and `numpy.int64` on everything else. _It looks like this was done on purpose_, but I can't figure out why. Maybe because I thought it should be distinct from `kLong64` and `kULong64`?

If something comes to you in a file, it must have a platform-independent type definition. I can think of no good reason for uproot to use `numpy.long` to interpret it.

Let's see if AppVeyor passes.

jpivarski:
@chrisburr It passed! Are you ready for me to merge this?

chrisburr:
Wonderful, I'm glad improving the conda-forge tests turned out to be useful! Thanks for figuring out the actual problem. You can merge whenever you're ready.

FYI: The update PRs on conda-forge now auto-merge provided the CI passes! 🎉

PR #449, opened 2020-02-11
Add support for streamer_aliases

### LINKED ISSUE #448: Name mismatch for PositionVector3D streamer
Yet another fun streamer naming thing... 😉

When opening a file from the CMS opendata, loading the `Tree` object fails with a `KeyError`. This seems to be because the expected key in `streamerinfosmap` is:

`b'ROOT::Math::PositionVector3D<ROOT::Math::Cartesian3D<double> >'`

I can work around it by hacking the key to be:

`b'ROOT::Math::PositionVector3D<ROOT::Math::Cartesian3D<double>,ROOT::Math::DefaultCoordinateSystemTag>'`

## Full reproducer
```python
In [1]: import uproot
   ...: f = uproot.open('root://eospublic.cern.ch//eos/opendata/cms/Run2010B/MuOnia/AOD/Apr21ReReco-v1/0000/02186E3C-D277-E011-8A05-00215E21D516.root')
   ...: tree = f['Events']
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-1-6b196f960e9d> in <module>
      1 import uproot
      2 f = uproot.open('root://eospublic.cern.ch//eos/opendata/cms/Run2010B/MuOnia/AOD/Apr21ReReco-v1/0000/02186E3C-D277-E011-8A05-00215E21D516.root')
----> 3 tree = f['Events']

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/rootio.py in __getitem__(self, name)
    223
    224     def __getitem__(self, name):
--> 225         return self.get(name)
    226
    227     def __len__(self):

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/rootio.py in get(self, name, cycle)
    357
    358             if last is not None:
--> 359                 return last.get()
    360             elif cycle is None:
    361                 raise _KeyError("not found: {0}\n in file: {1}".format(repr(name), self._context.sourcepath))

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/rootio.py in get(self, dismiss)
   1024
   1025         try:
-> 1026             return _classof(self._context, self._fClassName).read(self._source, self._cursor.copied(), self._context, self)
   1027         finally:
   1028             if dismiss:

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/rootio.py in read(cls, source, cursor, context, parent)
    964         out = cls.__new__(cls)
    965         out = cls._readinto(out, source, cursor, context, parent)
--> 966         out._postprocess(source, cursor, context, parent)
    967         return out
    968

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/tree.py in _postprocess(self, source, cursor, context, parent)
    287
    288         for branch in self._fBranches:
--> 289             self._attachstreamer(branch, context.streamerinfosmap.get(getattr(branch, "_fClassName", None), None), context.streamerinfosmap, False)
    290
    291         self._branchlookup = {}

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/tree.py in _attachstreamer(self, branch, streamer, streamerinfosmap, isTClonesArray)
    279                     pass
    280
--> 281                 self._attachstreamer(subbranch, submembers.get(name, None), streamerinfosmap, isTClonesArray)
    282
    283     def _postprocess(self, source, cursor, context, parent):

~/miniconda3/envs/d2hll/lib/python3.7/site-packages/uproot/tree.py in _attachstreamer(self, branch, streamer, streamerinfosmap, isTClonesArray)
    266                         base, name = name[:index], name[index + 1:]
    267                         if base in submembers and isinstance(submembers[base], digDeeperTypes):
--> 268                             submembers = streamerinfosmap[submembers[base]._fTypeName.rstrip(b"*")].members
    269
    270                 try:

KeyError: b'ROOT::Math::PositionVector3D<ROOT::Math::Cartesian3D<double> >'
```

chrisburr:
Fixes #448.

What do you think of this solution? I think it won't impact performance as it has no effect unless it until it crashes. (Well...technically it has to set up the try/except but that's effectively free in Python.)

### DIFF
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -418,3 +418,10 @@ class Test(object):
         assert 4 == len(a[0])
         with pytest.raises(IOError):
             len(b[0])
+
+    def test_issue448(self):
+        pytest.importorskip("pyxrootd")
+        f = uproot.open('root://eospublic.cern.ch//eos/opendata/cms/Run2010B/MuOnia/AOD/Apr21ReReco-v1/0000/02186E3C-D277-E011-8A05-00215E21D516.root')
+        tree = f['Events']
+        assert len(tree.arrays(entrystop=0)) == 4179
+        assert len(tree.arrays('recoMuons_muons__RECO.*', entrystop=10)) == 93
--- a/uproot/interp/auto.py
+++ b/uproot/interp/auto.py
@@ -558,3 +558,8 @@ interpret._itemdimpattern = re.compile(br"\[([1-9][0-9]*)\]")
 interpret._itemanypattern = re.compile(br"\[(.*)\]")
 interpret._vectorpointer = re.compile(br"vector\<([^<>]*)\*\>")
 interpret._pairsecond = re.compile(br"pair\<[^<>]*,(.*) \>")
+
+streamer_aliases = {
+    re.compile(b'(ROOT::Math::(?:PositionVector3D|DisplacementVector3D)<ROOT::Math::Cartesian3D<(?:[^>]+)>)\\s+(>)'):
+        b'\\1,ROOT::Math::DefaultCoordinateSystemTag\\2',
+}
--- a/uproot/interp/auto.py
+++ b/uproot/interp/auto.py
@@ -559,7 +559,7 @@ interpret._itemanypattern = re.compile(br"\[(.*)\]")
 interpret._vectorpointer = re.compile(br"vector\<([^<>]*)\*\>")
 interpret._pairsecond = re.compile(br"pair\<[^<>]*,(.*) \>")
 
-streamer_aliases = {
-    re.compile(b'(ROOT::Math::(?:PositionVector3D|DisplacementVector3D)<ROOT::Math::Cartesian3D<(?:[^>]+)>)\\s+(>)'):
-        b'\\1,ROOT::Math::DefaultCoordinateSystemTag\\2',
-}
+streamer_aliases = [
+    (re.compile(b'(ROOT::Math::(?:PositionVector3D|DisplacementVector3D)<ROOT::Math::Cartesian3D<(?:[^>,]+)>)\\s+(>)'),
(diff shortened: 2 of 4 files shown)


jpivarski:
Oh, I see—I hadn't considered the fact that normal streamers (all streamers observed by users willing to report issues in the past 3 years) don't even need to touch the `streamer_aliases`, and hence this has essentially zero cost for the vast majority of cases.

You're not using the dict as a dict. If you're always going to iterate over its `items`, then why not just store it as a list of regex-replacement pairs? Regexes are more powerful, but they can never take advantage of the hashmap aspect. We have two optimization points to choose from:

   * Make it more powerful and expressive with regexes, at the expense of always having to traverse the list (_O(n)_). This would be a good choice for a very regular set of cases to capture—it may be an infinite set, but expressible as regexes without a lot of `|` logic.
   * Make it less powerful but more scalable with a string → string hashmap. If the number of special cases grows to be large (but finite!), then it still scales well (_O(1)_).

Or both: a string → string hashmap that, when it fails, falls back to a list of regex-string pairs. But that's probably too complicated—we know from experience that this list is not going to grow quickly and we can always add the additional layer if it does. (Actually, if we see the list growing, we should ask ourselves if we're missing something that ought to be handled with a general rule, rather than exceptional cases.)

I like your idea of using regexes, but as a list of pairs, please.

When you're ready for merging, give me positive confirmation and I'll merge it. If you also want to fast-track it for deployment, update the uproot/version.py to 3.11.3 and I'll deploy it right after merging.

chrisburr:
I've just added a minor change to the regex to make it slightly less permissive for the numerical template type (`,` is no longer accepted).

Merge whenever you're happy. I've bumped the version number but I'm happy to wait while you consider the name in #450.

PR #450, opened 2020-02-11
Add "compress" extra to allow all compression libs to be easily installed

chrisburr:
I recently modified LHCb's [DIRAC](https://github.com/DIRACGrid/DIRAC/#dirac) extension to make it use `uproot` for extracting metadata from ROOT files. This has worked really well but declaring the dependency is a little awkward as it has to list each compression library individually and stay updated if new ones are ever added.

What do you think about adding another extra that allows people to depend on all of the decompression related dependencies using `pip install uproot[decompress]` or adding `uproot[decompress]` to their package's `setup.py`?

Alternative name suggestions for this extra are welcome.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -145,7 +145,8 @@ setup(name = "uproot",
       install_requires = ["numpy>=1.13.1", "awkward>=0.12.0,<1.0", "uproot-methods>=0.7.0", "cachetools"],
       setup_requires = ["pytest-runner"],
       extras_require = {
-          'testing': ["pytest>=3.9", "pkgconfig", "lz4", "zstandard", 'backports.lzma;python_version<"3.3"', "xxhash", "mock", "requests"],
+          "testing": ["pytest>=3.9", "pkgconfig", "lz4", "zstandard", 'backports.lzma;python_version<"3.3"', "xxhash", "mock", "requests"],
+          "decompression": ["lz4", "zstandard", 'backports.lzma;python_version<"3.3"', "xxhash"],
       },
       classifiers = [
           "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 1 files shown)


jpivarski:
I like this idea! I've seen other projects that use it (like TensorFlow including GPU backends or not), but I've never looked into how it was done.

   * What happens if a user installs `uproot`, then later decides to install `uproot[compress]`? It doesn't matter to me whether pip carefully installs the difference or redundantly re-downloads uproot, but at least it should leave the user with the same thing they would have gotten if they initially installed `uproot[compress]`.
   * What if they initially install `uproot[compress]`, then when they update, forget and do `pip install -U uproot` without the `[compress]`? Does it do something reasonable? (Not getting the latest compression libraries but getting the latest uproot would be one reasonable possibility.)
   * I'm trying to think of an easier to remember name—they shouldn't have to look up whether it's `[compress]`, `[compression]`, `[decomrpession]`, etc. This also affects uproot-writing, so leaving off the "de" is meaningful. (Also, some people say, "uncompress", though never "uncompression". I'm just trying to think of what will lead to the fewest human guessing-errors.)

chrisburr:
I've had a play inside virtual environments. If you want to try the command for installing extras with git exists but isn't obvious:
```
pip install 'git+https://github.com/chrisburr/uproot.git@add-compression-libs-extra#egg=uproot[compress]'
```

I've abused my admin powers to edit your comment to so I can give numbered responses:

1. It does the correct thing, only installs the extra libs, everything else is `Requirement already satisfied`.
2. Updates `uproot` but not the other libs.
3. Yeah I'm not a fan of this name either. I first called it `readall` but I think that's too ambiguous with remote reading libraries. I've changed it to `compress` for now.
4. Uninstalling behaves like usual for pip in that only `uproot` is removed and not `awkward` or anything else

chrisburr:
@jpivarski Any further thoughts on this?

jpivarski:
@chrisburr I didn't realize this was still waiting. It looks good to me. Should I merge it (i.e. you're not planning to do anything else)?

chrisburr:
If you're happy with the name `compress` go ahead.

PR #457, opened 2020-02-23
Limit docstrings to 80 characters. Fixes #456

### LINKED ISSUE #456: uproot docstrings should use hard wrapping
When typing help(some_uproot_object), the output looks weird even on wide screens because the lines are very long. Example:
```
Help on TTree in module uproot.rootio object:

class TTree(uproot.tree.TTreeMethods, TNamed, TAttLine, TAttFill, TAttMarker)
 |  Adds array reading methods to TTree objects that have been streamed from a ROOT file.
 |  
 |  - square brackets (``__getitem__``) returns a branch by name (see :py:meth:`get <uproot.tree.TTreeMethods.get>`).
 |  - the ``len`` function (``__len__``) returns the number of entries (same as ``numentries``).
 |  - iteration (``__iter__``) has no implementation. This is to avoid confusion between iterating over all branches (probably not what you want, but fitting the pattern set by :py:class:`ROOTDirectory <uproot.rootio.ROOTDirectory>` and ``dict``) and iterating over the data.
 |  
```
In a Jupyter notebook, this looks like:
```
Help on TTree in module uproot.rootio object:

class TTree(uproot.tree.TTreeMethods, TNamed, TAttLine, TAttFill, TAttMarker)
 |  Adds array reading methods to TTree objects that have been streamed from a ROOT file.
 |  
 |  - square brackets (``__getitem__``) returns a branch by name (see :py:meth:`get
<uproot.tree.TTreeMethods.get>`).
 |  - the ``len`` function (``__len__``) returns the number of entries (same as ``numentries``).
 |  - iteration (``__iter__``) has no implementation. This is to avoid confusion between iterating 
over all branches (probably not what you want, but fitting the pattern set by 
:py:class:`ROOTDirectory <uproot.rootio.ROOTDirectory>` and ``dict``) and iterating over the 
data.
 |  
```
For the help to print nicely, I think the line endings need to be inserted manually at X characters.

tamasgal:
I started to reformat the docstrings to limit them to 80 characters but since it's a bit tedious, I thought it would be better to ask what you think about (to avoid possible waste of time 😉 ).

I'd stick to 80, according to PEP 8.

I could not test the Sphinx output since `make html` in the `docs` folder produces the docs with the default theme and settings.

So the second question is: how to produce the docs on the local machine?

### DIFF
--- a/uproot/__init__.py
+++ b/uproot/__init__.py
@@ -115,7 +115,8 @@ Some, however, are useful:
 
 By default, local files are read as memory-mapped arrays. Change this by setting
 
-    open("...", localsource=lambda path: uproot.LocalSource(path, **uproot.LocalSource.defaults))
+    from uproot import FileSource
+    open("...", localsource=lambda path: FileSource(path, **FileSource.defaults))
 
 The same procedure sets options for uproot.XRootDSource and uproot.HTTPSource.
 """
--- a/uproot/__init__.py
+++ b/uproot/__init__.py
@@ -14,13 +14,16 @@ Open ROOT files with uproot.open (for reading) or uproot.create (for read-write)
     file = uproot.open("http://path/to/my/file.root")
     writeable = uproot.create("/new/local/file.root")
 
-These file objects act like dicts; get objects like TTrees from them with square brackets.
+These file objects act like dicts; get objects like TTrees from them with square
+brackets.
 
     tree = file["path/to/events"]
     tree = file["path/to/events;2"]            # optional cycle number
-    writeable["name"] = numpy.histogram(...)   # write to files by assignment (histograms only)
+    # write to files by assignment (histograms only)
+    writeable["name"] = numpy.histogram(...)
 
-TTree objects also act like dicts; get branches with square brackets or list them with keys().
+TTree objects also act like dicts; get branches with square brackets or list them
+with keys().
 
     tree.keys()
     tree.allkeys()    # recursive branches-of-branches
@@ -34,11 +37,13 @@ Get data as arrays with an array(...) or arrays(...) call.
     tree.arrays(["branch1", "branch2", "branch3"])
     tree.arrays(["Muon_*"])
 
-Variable numbers of objects per entry (particles per event) are handled by awkward-array:
+Variable numbers of objects per entry (particles per event) are handled by
+awkward-array:
 
     https://github.com/scikit-hep/awkward-array
 
-The arrays(...) call returns a dict from branch name (bytes) to data (Numpy array) by default.
+The arrays(...) call returns a dict from branch name (bytes) to data
+(Numpy array) by default.
 Change this by passing an outputtype class (e.g. dict, tuple, pandas.DataFrame).
 
     x, y, z = tree.arrays(["x", "y", "z"], outputtype=tuple)
@@ -47,8 +52,9 @@ For more idiomatic Pandas defaults, use tree.pandas.df().
 
     df = tree.pandas.df()
 
-If the desired branches do not fit into memory, iterate over chunks of entries with iterate().
-The interface is the same as above: you get the same dict/tuple/DataFrame with fewer entries.
+If the desired branches do not fit into memory, iterate over chunks of entries
+with iterate(). The interface is the same as above: you get the same
+dict/tuple/DataFrame with fewer entries.
 
     for x, y, z in tree.iterate(["x", "y", "z"], outputtype=tuple):
         do_something(x, y, z)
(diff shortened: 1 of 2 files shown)


jpivarski:
I'm in favor of reformatting the docstrings to 80 characters. I'm a little surprised that the `help` function doesn't do it automatically, considering all the other text precessing that the `help` function does.

About making it less tedious: doesn't Emacs's meta-Q hard-wrap the current paragraph to some width? Or, speaking of automated reformatters, what about writing a small function in the `uproot._help` sub-module that turns the last space before the 80th character into `"\n"` plus the number of leading spaces in the previous line (i.e. preserving indentation). Maybe something special would be needed for argument item lists (e.g. indent to colon?).

The advantage of this is to that we can change the parameters later.

About the Sphinx documentation, I usually make it locally with `make html`. The style is a little different from what readthedocs produces, but I've never tried to make them the same.

tamasgal:
Yes I tried to do some wrapping magic inside Emacs but could not make it work properly, with whatever plugins I have tried 😉

I think I like the programmatic approach idea, I'll have a look today!

chrisburr:
I recently used https://github.com/myint/docformatter to automatically tidy up the docstrings in a ~100,000 line project and it went well.

tamasgal:
Ah ok, I will look into this. Meanwhile I came up with this:

```python
import textwrap
import functools

def _join_and_preserve_sequential_indent(lines):
    if not lines:
        return ''
    first_line = lines[0]
    leading_spaces = len(first_line) - len(first_line.lstrip(' '))
    return '\n'.join([first_line] + [' ' * leading_spaces + line for line in lines[1:]])

def wrap(text, width=80):
    return "\n".join(
        map(join_and_preserve_sequential_indent,
            map(functools.partial(textwrap.wrap, width=width), text.split('\n'))))

text = u"""A :py:class:`Source <uproot.source.source.Source>` for compressed data.

    Decompresses on demand--- without caching the result--- so cache options in higher-level array functions are very important.

    Ordinary users would never create a :py:class:`CompressedSource <uproot.source.compressed.CompressedSource>`. They are produced when a TKey encounters a compressed value.

    Parameters
    ----------
    compression : :py:class:`Compression <uproot.source.compressed.Compression>`
        inherited description of the compression. Note that *this is overridden* by the first two bytes of the compressed block, which can disagree with the higher-level description and takes precedence.

    source : :py:class:`Source <uproot.source.source.Source>`
        the source in which compressed data may be found.

    cursor : :py:class:`Cursor <uproot.source.cursor.Cursor>`
        location in the source.

    compressedbytes : int
        number of bytes after compression.

    uncompressedbytes : int
        number of bytes before compression.
"""

print(wrap(text))
```

Which produces this:

```
A :py:class:`Source <uproot.source.source.Source>` for compressed data.

    Decompresses on demand--- without caching the result--- so cache options in
    higher-level array functions are very important.

    Ordinary users would never create a :py:class:`CompressedSource
    <uproot.source.compressed.CompressedSource>`. They are produced when a TKey
    encounters a compressed value.

    Parameters
    ----------
    compression : :py:class:`Compression <uproot.source.compressed.Compression>`
        inherited description of the compression. Note that *this is overridden*
        by the first two bytes of the compressed block, which can disagree with the
        higher-level description and takes precedence.

    source : :py:class:`Source <uproot.source.source.Source>`
        the source in which compressed data may be found.

    cursor : :py:class:`Cursor <uproot.source.cursor.Cursor>`
        location in the source.

    compressedbytes : int
        number of bytes after compression.

    uncompressedbytes : int
        number of bytes before compression.
```

tamasgal:
> I recently used https://github.com/myint/docformatter to automatically tidy up the docstrings in a ~100,000 line project and it went well.

Just had a quick look; I think the main problem is that the way the documentation is done in `uproot` is not "standard", meaning that fragments are declared in `_help` and injected (and reused) from that module to other classes. I have not tried the `docformatter` yet, but I suspect that it will not be able to pick up the stuff in `_help`, which is the major part of the documentation.

jpivarski:
The reason for that was because so many of the functions have identical arguments with identical meanings, and I wanted to use variable substitutions to ensure consistency. The docstrings in their standard locations have to be constant strings (in some versions of Python, at least).

That's another argument for a wrapping function: it can be applied dynamically, after the variable substitutions.

I like what your `_join_and_preserve_sequential_indent` is doing, and particularly the fact that it reuses Python's own `textwrap`. (That's the module that I had _thought_ was being applied to `help` output.)

I'm aware that I need to start breaking lines at 80 characters because that's a standard. In fact, I'm planning on doing it in one big sweep on Awkward 1, when it's "done." (I added all these uproot docstrings in one big sweep, because that way I could better ensure consistency.) Personally, I find the wrapped lines less flexible because your window has to be very close to 80 characters wise to make use of text formatted that way: a little narrower and it's impossible to read (every other line has a dribble of soft-wrapping); any wider and you're just wasting screen space (because it doesn't reflow). In April, I'll be updating Uproot to use Awkward 1, and I can do an 80-character wrapping as part of that sweep. (Not including this help text because we have your new wrapping function, unless the wrapping function also removes non-meaningful carriage returns, like markdown.)

tamasgal:
Yes, I was aware of that design decision (reusable fragments) and I like it.

I think introducing Awkward 1 in combination of a huge code style refactoring might be a bit too much, in case we need to debug regressions, but not sure. Anyways, I am really looking forward to Awkward 1!

I just updated the PR to use that `wrap` function. It seems to work nicely but let me know what you think.

jpivarski:
Okay, then I won't do the 80-character sweep in Uproot at the same time as the Awkward transition, but will consider it for sometime thereafter.

tamasgal:
@jpivarski do you want any further changes? Otherwise I am done with this.

jpivarski:
Looks great, thanks!

Issue #447, opened 2020-01-29
uproot.rootio.TTree.arrays omitting variables


kreczko:
### Summary
In the case of multiple branches with identical variable (sub-branch/leaf?) names the [`uproot.rootio.TTree.arrays`](https://github.com/scikit-hep/uproot/blob/master/uproot/tree.py#L502) will hide variables.

As en example, see http://fast-hep-data.web.cern.ch/fast-hep-data/cms/L1T/CMS_L1T_study.root where the tree `'l1CaloTowerEmuTree/L1CaloTowerTree'` has the structure:

```bash
CaloTP                     TStreamerInfo              None
nHCALTP                    TStreamerBasicType         asdtype('>i2')
hcalTPieta                 TStreamerSTL               asjagged(asdtype('>i2'), 10)
hcalTPiphi                 TStreamerSTL               asjagged(asdtype('>i2'), 10)
hcalTPCaliphi              TStreamerSTL               asjagged(asdtype('>i2'), 10)
hcalTPet                   TStreamerSTL               asjagged(asdtype('>f4'), 10)
hcalTPcompEt               TStreamerSTL               asjagged(asdtype('>i2'), 10)
hcalTPfineGrain            TStreamerSTL               asjagged(asdtype('>i2'), 10)
nECALTP                    TStreamerBasicType         asdtype('>i2')
ecalTPieta                 TStreamerSTL               asjagged(asdtype('>i2'), 10)
ecalTPiphi                 TStreamerSTL               asjagged(asdtype('>i2'), 10)
ecalTPCaliphi              TStreamerSTL               asjagged(asdtype('>i2'), 10)
ecalTPet                   TStreamerSTL               asjagged(asdtype('>f4'), 10)
ecalTPcompEt               TStreamerSTL               asjagged(asdtype('>i2'), 10)
ecalTPfineGrain            TStreamerSTL               asjagged(asdtype('>i2'), 10)

L1CaloTower                TStreamerInfo              None
nTower                     TStreamerBasicType         asdtype('>i2')
ieta                       TStreamerSTL               asjagged(asdtype('>i2'), 10)
iphi                       TStreamerSTL               asjagged(asdtype('>i2'), 10)
iet                        TStreamerSTL               asjagged(asdtype('>i2'), 10)
iem                        TStreamerSTL               asjagged(asdtype('>i2'), 10)
ihad                       TStreamerSTL               asjagged(asdtype('>i2'), 10)
[...]

Issue #444, opened 2020-01-27
Extend branch lookup to match subbranches using `/` as separator


ast0815:
As discussed in #443, it would be convenient to specify subbranches in the `iterate` functions. This would have to be implemented in a couple of places:

https://github.com/scikit-hep/uproot/blob/master/uproot/rootio.py#L333-L363

https://github.com/scikit-hep/uproot/blob/master/uproot/tree.py#L374-L383

https://github.com/scikit-hep/uproot/blob/master/uproot/tree.py#L981-L986

jpivarski:
The third part of [this comment](https://github.com/scikit-hep/uproot/issues/443#issuecomment-578831373).

ast0815:
So, I finally got around to look into this. It seems like the `ROOTDirectory.get` method already supports arbitrary '/' depths. So no more work needed there, it seems.

I got a few questions about the intended behaviour for the `TTree` objects though:

1)  Should `tree.iterkeys(recursive=True)` return the subkeys with the '/' separators or just the raw subkeys?
2)  Should the `name` properties of the branches returned by `tree.itervalues(recursive=True)` be modified to include the '/' or remain raw?
3)  Same questions for `tree.iteritems(recursive=True)`.
4) What should happen when `get` is called with `recursive=True` and a name containing a '/'?
    Or with '/' and `resursive=False` for that matter.
5) Should the default value of get really be `recursive=True`?

If it were up to me, I would probably say: modify the values of `iterkeys` and the key part of `iteritems` to contain the full path, but leave the `name` property of the returned values alone. This way each subbranch always gets a unique key as seen from the particular top-level tree, but the branches themselves remain untouched.

I also would remove the `recursive` argument from get and have it only look at exact matches with the `iterkeys` (It can decide whether to use `iterkeys(recursive=True/False` by checking whether there is a `/` in the requested name). This way there are no surprises and you can simply use `get` or `__getitem__` with the return values of `keys` or `allkeys` without thinking about it too much.

I'm not 100% sure, but this might also solve #447 in one swoop.

Thoughts?

jpivarski:
Your ideas sound good to me:

   * `iterkeys(recursive="/")` and `iteritems(recursive="/")` should yield keys that have the parent names prepended with slashes; that is, "fully qualified" keys,
   * these fully qualified keys should be usable in `get` (and therefore `__getitem__`), and
   * the `value.name` should _not_ be affected. Iterating over the keys and/or values should not cause data to change.

[...]

PR #453, opened 2020-02-17
Fixing issue #447: uproot.rootio.TTree.arrays omitting variables

kreczko:
Fixes #447 

 - added test & test data for issue #447
 - added `recursive` parameter to `arrays` method
 - added `_provenance` information to branches tracking which tree and parent branch(es) they are coming from
 - dictionary keys use `provenance` information for creation using `recursive` as the delimiter

This is probably not the most ideal solution - feedback is welcomed.

PR #459, opened 2020-02-29
Implement #444

ast0815:
I have implemented the changes discussed in #444. The tests currently fail, because this branch is based on @kreczko's branch for fixing #447. It uses the same sample root file for the tests. I did not touch the test for #447, which is why that still fails.

### DIFF
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -454,10 +454,21 @@ class Test(object):
             assert sorted(list(arrays.keys())) == expectedKeys
 
     def test_issue444_subbranche_lookup_with_slash():
-        # Access subbranches directly
         with uproot.open("tests/samples/issue447.root") as f:
+            # Access subbranches directly from file
             assert numpy.all(f['l1CaloTowerEmuTree/L1CaloTowerTree/CaloTP']['nECALTP'].array()
                 == f['l1CaloTowerEmuTree/L1CaloTowerTree/CaloTP/nECALTP'].array())
+            # Access subbranches from TTree
+            tree = f['l1CaloTowerEmuTree/L1CaloTowerTree']
+            assert numpy.all(tree['CaloTP']['nECALTP'].array()
+                == tree['CaloTP/nECALTP'].array())
+            # Test different recursive schemes
+            assert 'CaloTP/nECALTP' in tree.keys(recursive='/')
+            assert 'CaloTP/nECALTP' not in tree.keys(recursive=True)
+            assert 'CaloTP/nECALTP' not in tree.keys(recursive=False)
+            assert 'nECALTP' not in tree.keys(recursive='/')
+            assert 'nECALTP' in tree.keys(recursive=True)
+            assert 'nECALTP' not in tree.keys(recursive=False)
         # Specify subbranches in iterate
         for arrays in uproot.iterate(["tests/samples/issue447.root"], 'l1CaloTowerEmuTree/L1CaloTowerTree', ['CaloTP/nECALTP']):
             pass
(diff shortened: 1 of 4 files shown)


jpivarski:
@kreczko, does any of this duplicate or conflict with your PR? Can we merge the efforts?

ast0815:
I just rebased my changes to start on the correct commit from @kreczko's pull request.

kreczko:
hmm..it is not easy to see. There are definitely related elements, but the use case is different.

`_provenance` from MR #459 can definitely be used to construct the `iteritems` and is probably doable for sub-branches.

We used to do something similar but would simply generate additional aliases for sub-branch access on top of uproot (wrapping uproot and directly mapping to provided numpy arrays):

```python

# class providing recursive mapping
class MyAliasTree(dict):
  ...

t = MyAliasTree(filename)
t['l1CaloTowerEmuTree/L1CaloTowerTree/CaloTP/nECALTP'] == t['l1CaloTowerEmuTree/L1CaloTowerTree']['CaloTP/nECALTP'] ==
t['l1CaloTowerEmuTree/L1CaloTowerTree']['CaloTP']['nECALTP']
```

ast0815:
I'm not sure additional aliases would solve this. If you have multiple levels of nested trees, then each (sub-)^n branch can have n different aliases depending on what tree level you are asking. Also the aliases for subbranches on different subtrees might be identical as seen from their respective subtree, so when questioning the base tree you get chances for confusion again. Or maybe I am misunderstanding the suggestion.

ast0815:
Is there a planned way forward with this? Anything I could do?

jpivarski:
This has been in limbo for a while—I just noticed it when scanning through old issues.

This feature, as I understand it, exists in Uproot4 (much easier to write it anew than change the assumptions of an existing codebase). If what you see there is what you were wanting, then there may be nothing to do. If you try it out (in Uproot4) and find that it's not quite right, let me know how it could be better.

In either case, let me know if you want to keep this PR open or to retire it.

ast0815:
I'll have another look at some point and probably rebase this now that #453 is merged.

jpivarski:
Considering that Uproot 4 will be replacing Uproot 3 this year, the higher priority would be to be sure that the feature exists in Uproot 4. Since you're more familiar with what this is supposed to do, could you determine whether Uproot 4's implementation satisfies the requirement? (It was about slashes in key strings, right? I made sure to include that in the new version, but is it done right, as you would want it?)

ast0815:
I'll give it a look. 👍

ast0815:
Hello again,

I just checked and it seems like addressing subbranches from file and tree level with '/' as level separator works as intended in uproot4. I could not find any `up.iterate` or `up.pandas.iterate` function though, so I could not test whether it supports the use case that originally prompted me to find the issue. I.e. I wanted to iterate over many root files and get pandas DataFrames with columns from nested branches when the explicit branch name was not unique.

jpivarski:
I've been focusing on deserializing all the data types, but since I'll want to present it soon, I was planning on adding `iterate` next. (Maybe tomorrow, after porting all the old `test_issues.py`.) So you're right that `uproot.iterate` isn't there yet.

As part of a general campaign of unifying things, `uproot.pandas.iterate` is going to be `library="pd"` on `uproot.iterate`. The `library` parameter determines the output library; `library="np"` means NumPy (even if jagged: we make NumPy object arrays if need be), `library="ak"` is Awkward, `library="pd"` is Pandas (Series and DataFrames), and `library="cp"` is CuPy (GPUs). Others may be added in the future.

ast0815:
Anything else I should do for this to be merged?

jpivarski:
That works! Thanks!

PR #463, opened 2020-03-05
Comment out redundant line in TFileUpdate

### LINKED ISSUE #460: uproot.update is not functional
If you look at the lines here

https://github.com/scikit-hep/uproot/blob/6b7e886efc0720a20b3622a5b2840cb1fa0558bc/uproot/write/TFile.py#L27-L40

running `uproot.update` on an existing ROOT file causes a crash like so:

```
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "run.py", line 13, in process_file
    with uproot.update(output_fname) as out_file:
  File "/afs/cern.ch/user/g/gstark/pyhf/lib64/python3.6/site-packages/uproot/write/TFile.py", line 29, in __init__
    self._openfile(path)
TypeError: _openfile() missing 1 required positional argument: 'compression'
```

this is because the call to `_openfile` in the `__init__` seems to only specify one positional argument here.

reikdas:
Fixes #460

### DIFF
--- a/uproot/write/TFile.py
+++ b/uproot/write/TFile.py
@@ -26,7 +26,8 @@ from uproot.write.objects.TTree import TTree
 
 class TFileUpdate(object):
     def __init__(self, path):
-        self._openfile(path)
+        #FIXME: Need to fix self.compression attribute
+        #self._openfile(path)
         raise NotImplementedError
 
     def _openfile(self, path, compression):


jpivarski:
The failures seem to be because ROOT isn't being installed. Do you see any reason why that could be? If not, maybe we could try to reproduce it locally or ask Chris to take a look.

reikdas:
ROOT is being installed, but RDataFrame is not being able to import Numpy properly. (But this is not a ROOT 6.20 error - the version on conda, because the tests pass when using my locally built ROOT). 
I am not sure what is going wrong :( 

@chrisburr any ideas?

chrisburr:
There is a bug in how 6.20.0 has been packaged for conda-forge.

I've fixed it in https://github.com/conda-forge/root-feedstock/pull/79 and it should be available in a few hours.

Thanks for letting me know!

chrisburr:
I've restarted the CI and it's now fixed.

reikdas:
Thanks @chrisburr :D 

@jpivarski Can we merge this now?

PR #471, opened 2020-04-03
Read RNTuple anchor

jpivarski:


jpivarski:
Commit 317c6c44e5348145ad288d93068efc19d2c79cf7 was unintentionally added, but it's not a problem. We wanted to bump the version number anyway.

jpivarski:
@oshadura This PR passes tests, but I can keep it open in case you run into anything else that needs to be fixed at the uproot level, anything that could help you figure out how to deserialize RNTuple in Python. If, on the other hand, you think this is all that's necessary, I can merge it so that you can work from a released version.

oshadura:
Hi Jim! I realized that the easiest way is to make a rntuple reader as a dependency of uproot, but I can't proceed without your PR :) (particularly without rntuple defined in `builtin_classes`).

jpivarski:
I can't believe I forgot to merge this. I'll also deploy a version soon.

oshadura:
Wonderfu, now I can go ahead full speed!

PR #476, opened 2020-04-22
[WIP] implementation of s3 protocol

wiso:
Implement S3 object storage. S3 object storage is provided by aws but it is also some universities have their own object storage based on S3 (e.g. using CEPH).

ROOT support reading from s3 thanks to [TS3WebFile](https://root.cern.ch/doc/master/classTS3WebFile.html)

The implementation is based on the [s3fs](https://s3fs.readthedocs.io/en/latest/index.html) python packages, which is based on [boto](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html).

The implementation is very basic, copying the one from source/file.py.

There is also a new test case, but I am not sure how to run it in general since it needs authentication (by the way it works).

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -1,0 +1,20 @@
+import pytest
+
+import uproot
+
+
+# export S3_SECRET_ACCESS_KEY=
+# export S3_ACCESS_KEY_ID=
+
+# use these by now
+# export AWS_ACCESS_KEY_ID=
+# export AWS_SECRET_ACCESS_KEY=
+
+URL = "s3http://rgw.fisica.unimi.it/test-ruggero/test.root"
+
+class Test(object):
+    def test_s3(self):
+        f = uproot.open(URL)
+        assert type(f) == uproot.rootio.ROOTDirectory
+        histo = f.get("H")
+        assert(histo.name == b'H')
--- a/uproot/rootio.py
+++ b/uproot/rootio.py
@@ -23,12 +23,13 @@ from uproot.source.memmap import MemmapSource
 from uproot.source.xrootd import XRootDSource
 from uproot.source.http import HTTPSource
 from uproot.source.cursor import Cursor
+from uproot.source.s3 import S3Source
 
 import uproot_methods.classes
 
 ################################################################ high-level interface
 
-def open(path, localsource=MemmapSource.defaults, xrootdsource=XRootDSource.defaults, httpsource=HTTPSource.defaults, **options):
+def open(path, localsource=MemmapSource.defaults, xrootdsource=XRootDSource.defaults, httpsource=HTTPSource.defaults, s3source=S3Source.defaults, **options):
     if isinstance(path, getattr(os, "PathLike", ())):
         path = os.fspath(path)
     elif hasattr(path, "__fspath__"):
@@ -59,6 +60,9 @@ def open(path, localsource=MemmapSource.defaults, xrootdsource=XRootDSource.defa
     elif _bytesid(parsed.scheme) == b"http" or _bytesid(parsed.scheme) == b"https":
         return http(path, httpsource=httpsource, **options)
 
+    elif _bytesid(parsed.scheme) == b"s3http":
+        return s3http(path, s3source=s3source, **options)
+
     else:
         raise ValueError("URI scheme not recognized: {0}".format(path))
 
@@ -87,6 +91,19 @@ def http(path, httpsource=HTTPSource.defaults, **options):
     else:
         openfcn = httpsource
     return ROOTDirectory.read(openfcn(path), **options)
+
+def s3http(path, s3source=S3Source.defaults, **options):
+    if isinstance(s3source, dict):
+        kwargs = dict(S3Source.defaults)
+        kwargs.update(s3source)
+        for n in kwargs:
+            if n in options:
+                kwargs[n] = options.pop(n)
+        openfcn = lambda path: S3Source(path, **kwargs)
+    else:
+        openfcn = s3source
+    return ROOTDirectory.read(openfcn(path), **options)
+
 
 def nofilter(x): return True
 
--- a/uproot/source/s3.py
+++ b/uproot/source/s3.py
@@ -1,0 +1,59 @@
+from __future__ import absolute_import
+
+import multiprocessing
+import sys
+from urllib.parse import urlparse
+
+import numpy
+
+import uproot.source.chunked
+
+
+class S3Source(uproot.source.chunked.ChunkedSource):
+    # makes __doc__ attribute mutable before Python 3.3
+    __metaclass__ = type.__new__(type, "type", (uproot.source.chunked.ChunkedSource.__metaclass__,), {})
+
+    def __init__(self, path, *args, **kwds):
+        super(S3Source, self).__init__(path, *args, **kwds)
+
+        url_parsed = urlparse(self.path)
+        if "s3" not in url_parsed.scheme:
+            raise ValueError("%s not a valid s3 url" % self.path)
+        scheme = url_parsed.scheme.replace("s3", "")
+        self.endpoint_url = scheme + "://" + url_parsed.netloc
+        self.local_path = url_parsed.path
+        if not self.local_path:
+            raise ValueError("%s not a valid s3 url" % self.path)
+
+        self.fs = None  # the filesystem
+        self._size = None
+
+    defaults = {
+        "chunkbytes": 1024 ** 2,
+        "limitbytes": 100 * 1024 ** 2,
+        "parallel": 8 * multiprocessing.cpu_count() if sys.version_info[0] > 2 else 1,
+    }
+
+    def size(self):
+        if self._size is None:
+            self._open()
+        return self._size
+
+    def _open(self):
+        try:
+            import s3fs
+        except ImportError:
+            raise ImportError("Install s3fs package with:\n    conda install -c conda-forge s3fs.")
+        if self.fs is None:
+            self.fs = s3fs.S3FileSystem(client_kwargs={"endpoint_url": self.endpoint_url})
+        self._source = self.fs.open(self.local_path, "rb")
+        self._size = self._source.info()["size"]
+
+    def _read(self, chunkindex):
+        self._source.seek(chunkindex * self._chunkbytes)
+        data = self._source.read(self._chunkbytes)
+        return numpy.frombuffer(data, dtype=numpy.uint8)
+
+    def dismiss(self):
+        if self._source is not None:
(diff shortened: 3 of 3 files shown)


chrisburr:
Is there a reason why this needs to be a dedicated protocol? I think it be better to [sign the URL](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html) and use the normal http(s) source for reading.

jpivarski:
> Is there a reason why this needs to be a dedicated protocol?

It's not going through Uproot's HTTP method; it uses a package called `s3fs`, which behaves like a file handle at the Python level.

wiso:
> This implementation looks right—I don't see any red flags and I'd accept it. As-is, I don't think it would pass tests because Travis probably didn't have s3fs installed, but we could include that in the .travis.yml, which is conda-based.

I guess several things should be fixes, for example I have no idea how `threadlocal` method works, but I guess it should be implemented.

> 
> Tying the tests to a file that you host is more of an issue. You certainly wouldn't want to put your credentials into the test suite. Perhaps this needs to be something that is only conformably tested?

I don't have a better solution...

> Is the appropriate protocol "s3http"? I have seen `s3://` elsewhere.

I am using the same convenction of ROOT, see from https://root.cern.ch/doc/master/classTS3WebFile.html#ac3e4f2023440a3846b1360fa271c876e:

```
The specified scheme (i.e. s3, s3http, s3https, ...) determines the underlying transport protocol to use for downloading the file contents, namely HTTP or HTTPS. The 's3', 's3https', 'gs' and 'gshttps' schemes imply using HTTPS as the transport protocol. The 's3http', 'as3' and 'gshttp' schemes imply using HTTP as the transport protocol.
```

chrisburr:
> It's not going through Uproot's HTTP method; it uses a package called s3fs, which behaves like a file handle at the Python level.

I realise, I'm just not convinced there is anything to to gain from using `s3fs` instead of `requests` (behind the scenes it will be http(s) regardless).

Performance has the potential to be better when I find time to implement https://github.com/scikit-hep/uproot/issues/393 whereas `s3fs` would require a dedicated implementation like how XRootD will.

jpivarski:
[Amazon's documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html) describes access through HTTP(S) and the S3:// protocol (section titled "Accessing a Bucket using S3://").

If this can be a special case of HTTP(S)/requests, I'd prefer that. It wouldn't require any additional dependencies and Chris has already done some optimization of HTTP(S).

If s3fs goes through unencrypted HTTP (the reason you're not using "s3://"), then that would be another point in favor of using HTTPSource. We may be able to find a programmatic way to pass credentials to AWS, too, rather than relying on environment variables (fine for a one-user script, becomes hard if used as an infrastructure component).

wiso:
> > It's not going through Uproot's HTTP method; it uses a package called s3fs, which behaves like a file handle at the Python level.
> 
> I realise, I'm just not convinced there is anything to to gain from using `s3fs` instead of `requests` (behind the scenes it will be http(s) regardless).
> 
> Performance has the potential to be better when I find time to implement #393 whereas `s3fs` would require a dedicated implementation like how XRootD will.

Using requests on top of a http url would be very good. I tried the method you linked and I am able to generate a http url.
The question is: do we prefer the user to do this translation (s3 -> http) or do you want it to be inside uproot?

jpivarski:
If there's a programmatic mapping from `s3://` URLs to `https://` URLs, then it would be nice to encode that knowledge in

   * [uproot.open](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/rootio.py#L31-L63)
   * [uproot.iterate](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L115-L154)
   * [uproot.numentries](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L2048-L2053)
   * [uproot.pandas.iterate](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/pandas.py#L13-L15)

by calling a centralized function to translate `urllib.parse.ParseResult` objects into new `urllib.parse.ParseResult` objects from `s3://` to their equivalent `https://`, passing all others through untouched.

Actually, the only places where `urlparse` is called are [here](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/rootio.py#L41) and [here](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L70), so this translator only needs to be called at two sites to update all four functions that take URLs.

wiso:
> If there's a programmatic mapping from `s3://` URLs to `https://` URLs, then it would be nice to encode that knowledge in
> 
>     * [uproot.open](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/rootio.py#L31-L63)
> 
>     * [uproot.iterate](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L115-L154)
> 
>     * [uproot.numentries](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L2048-L2053)
> 
>     * [uproot.pandas.iterate](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/pandas.py#L13-L15)
> 
> 
> by calling a centralized function to translate `urllib.parse.ParseResult` objects into new `urllib.parse.ParseResult` objects from `s3://` to their equivalent `https://`, passing all others through untouched.
> 
> Actually, the only places where `urlparse` is called are [here](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/rootio.py#L41) and [here](https://github.com/scikit-hep/uproot/blob/515b219dda389e6257b5611cb5b9844102236853/uproot/tree.py#L70), so this translator only needs to be called at two sites to update all four functions that take URLs.

Yes, this is possible, but still needs external libraries (boto3, see the link from @chrisburr ) and only the one who has access to the s3 resource can make this mapping.

jpivarski:
> Yes, this is possible, but still needs external libraries (boto3, see the link from @chrisburr ) and only the one who has access to the s3 resource can make this mapping.

By the mapping, I was referring to a string-to-string (or maybe a `urllib.parse.ParseResult` to `urllib.parse.ParseResult`) conversion, which ought to be simple enough that it doesn't require a library. The four URL-taking functions might need a few additional arguments for username and credentials.

Isn't it something like replacing `s3://XYZ/file/path.root` with `https://XYZ.s3.amazon.com/file.path.root` (+ HTTPS authentication stuff)? It's in AWS's interests to make it simple enough for developers to adhere to a protocol without needing a library.

jpivarski:
Oh, I see: I just caught your reference to Chris's link about generating presigned URLs. That's probably a REST interface without strict library requirements as well, but it is an additional complication.

chrisburr:
> Oh, I see: I just caught your reference to Chris's link about generating presigned URLs. That's probably a REST interface without strict library requirements as well, but it is an additional complication.

Signing a URL is a fully offline operation. You basically hash the URL with your credentials and an expiry timestamp. You can then safely share it without exposing the secret and it is then just a standard URL that can be used by any piece of software. The sever then does the same hashing operation with the shared secret to see if the signature is valid. It's quite neat as a decentralised authentication mechanism.

If this becomes a feature of uproot it is probably preferable to use the standard way of authenticating using a header containing a hash. This is what [ROOT does for `TS3WebFile`](https://github.com/root-project/root/blob/master/net/net/src/TS3HTTPRequest.cxx) by inheriting from `TWebFile` just overloading the function that makes the request.

chrisburr:
> Yes, this is possible, but still needs external libraries (boto3, see the link from @chrisburr ) and only the one who has access to the s3 resource can make this mapping.

Sorry, foolishly working though my email backwards, so only just saw this.

It's not essential to use `boto3` (though I normally do). Amazon provides a Python example that only uses the standard library: https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html

These two pages explain exactly what is needed for each type of request: http://awsdocs.s3.amazonaws.com/S3/latest/s3-qrc.pdf

wiso:
I have thought a bit and now I am not sure if we want this feature in uproot. The workflow may be:

1. the user has a ROOT file in a s3 bucket
2. the user convert the s3 url (with boto, with s3 cli, ...) into a signed http url
3. use it with uproot

If otherwise you prefer to have uproot able to handle s3 url directly (doing the sign internally) we have to agree on some details:

1. presign or sign? The first link @chrisburr sent (boto3) was about pre-sign, while the python example about sign. I am not sure about the difference, but with pre-sign it should be possible also to upload a file, not sure if we care
2. What kind of request to send? GET + authentication in the header, POST, GET with authentication in the query. Seems ROOT is using the first but it should be no big difference

I see that ROOT support both amazon s3 and google cloud storage. The signature algorithm is very similar, but maintaining everything could be quite annoying, in particular if one these service change the signature method.

chrisburr:
>     1. presign or sign? The first link @chrisburr sent (boto3) was about pre-sign, while the python example about sign. I am not sure about the difference, but with pre-sign it should be possible also to upload a file, not sure if we care

Normal signing is better if it's a feature of uproot. Presigned URLs are good for providing access to read/create an object without needing explicit support but S3, like how can be done with the current release of uproot, but it does have limitations.

Good remote writing is support is impossible with presigned URLs as you would need keep getting new URLs for each request you make for modifying each part of the file (there is also a 5GB limit per request and multi-chunk uploads require many different URLs to be signed). Long lived objects might also have issues as presigned URLs have an expiry time associated with them.

jpivarski:
> Good remote writing is support is impossible...

Don't worry about writing. Uproot's file-writing code would have to be significantly rethought if the output file were remote. There's a lot of backtracking it has to do to keep the file consistent as histograms and TTree baskets are incrementally added. The current implementation assumes that we can update any byte range in place without a performance penalty.

As for Uproot and S3 more generally, I'll support whichever choice you make: if it's easier for your workflow for Uproot to know some S3 details and handle S3 URLs internally, then let's add it, but if signing an HTTP URL and using Uproot's built in HTTP support makes more sense, then this should be a how-to tutorial.

You already have a question open on StackOverflow: perhaps you could answer your own question with a recipe and I'll either delete mine or we'll upvote yours past mine. You're allowed to select your own answer as "the" answer to a question.

wiso:
Sorry for the delay, I think using http is probably the best. I have update stackoverflow.

PR #493, opened 2020-06-03
Fixing bugs for HATS.

jpivarski:
The first of these is a long-held oversight: Uproot has never been able to read @sbinet's small-evnt-tree-nosplit.root. Unsplit objects are much harder, so maybe this was justified in the beginning, but now I wanted to show the same example in split and unsplit form, for which small-evnt-tree-fullsplit.root and small-evnt-tree-nosplit.root provide a good example.

The two bugs that had to be fixed were:

   * Python's `l` and `L` don't correspond to `long` and `unsigned long`; they're both 32-bits. Different platforms make similarly different interpretations. For 64-bit non-Windows, at least, the data should be deserialized as 64-bit, so we use Python's `q` and `Q`.
   * `std::vector<std::string>` wasn't implemented for unsplit objects. The implementation is easy.


jpivarski:
The Uproot HATS is done, so I'll close off this PR and give it a version number.

Issue #507, opened 2020-07-11
Unable to read vector<long long>/vector<Long64_t> branches


masonproffitt:
The type `std::vector<long long>` (or `std::vector<Long64_t>`) is missing from uproot's type interpretation:

```python
>>> import uproot
>>> file = uproot.open('test_file.root')
>>> file['tree']['long64_t_vector_branch']
<TBranchElement b'long64_t_vector_branch' at 0x7f87add58250>
>>> file['tree']['long64_t_vector_branch'].array()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/site-packages/uproot/tree.py", line 1362, in array
    interpretation = self._normalize_interpretation(interpretation, awkward)
  File "/usr/lib/python3.8/site-packages/uproot/tree.py", line 1105, in _normalize_interpretation
    raise ValueError("cannot interpret branch {0} as a Python type\n   in file: {1}".format(repr(self.name), self._context.sourcepath))
ValueError: cannot interpret branch b'long64_t_vector_branch' as a Python type
   in file: /home/user/cern/iris/code/uproot_tree_utils/test_file.root
```

The type is missing in `uproot/interp/auto.py`. `unsigned long long`/`ULong64_t` works fine.

PR #508, opened 2020-07-11
Add interpretation for vector<long long>

masonproffitt:
Addresses #507. By https://root.cern.ch/doc/master/RtypesCore_8h.html, `Long_t` is `long` and `Long64_t` is `long long` (and similarly for the unsigned versions).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
 
   - PYVER=3.8 NPY="numpy>=1.15"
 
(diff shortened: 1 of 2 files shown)


jpivarski:
Thanks—I should include this in Uproot 4, too. Fortunately, there's only one place to put it in Uproot 4 (in a real parser... well, tokenizer and set of passing rules).

jpivarski:
Sorry that this took so long to fix and merge!

PR #514, opened 2020-08-13
NumPy's tostring -> tobytes warning (Python 3.8 stops suppressing deprecation warnings)

EdoPro98:
Changed deprecated method `tostring()` with method `tobytes()` for compatibility with python versions newer than 3.8

### DIFF
--- a/uproot/_util.py
+++ b/uproot/_util.py
@@ -1,0 +1,11 @@
+#!/usr/bin/env python
+
+# BSD 3-Clause License; see https://github.com/scikit-hep/uproot/blob/master/LICENSE
+
+from __future__ import absolute_import
+
+def _tobytes(x):
+    if hasattr(x, "tobytes"):
+        return x.tobytes()
+    else:
+        return x.tostring()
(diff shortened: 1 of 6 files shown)


jpivarski:
Thanks for this, though it's not so much a Python 3.8 thing as a recent NumPy thing, and we do support NumPy from 1.13.1 up, which don't have the `tobytes` method (it would `AttributeError`. What I did in scikit-hep/uproot4#63 is to add a utility function that calls `tobytes` if it exists, `tostring` if it doesn't:

https://github.com/scikit-hep/uproot4/pull/63/files

You or I can modify this PR to switch between `tobytes` and `tostring`, but it will have to be done before I can merge it (lest we break many user's workflows!). Thanks!

jpivarski:
Thanks for starting this! If I don't hear back, I'll merge it after the tests pass.

(My change checks NumPy arrays for the existence of `tobytes` and uses it if it exists; otherwise, we use `tostring`. Switching based on Python version wouldn't catch cases in which someone is using old NumPy with new Python. "Old NumPy" in this case is only a few years old.)

PR #517, opened 2020-08-24
Update tree.py

JMSchoeffmann:
Problem in VS Code, Win 10, Python 3:
pattern = os.path.expanduser(path)
UnboundLocalError: local variable 'path' referenced before assignment
My Change worked fine in my case, don't know for everyone.


jpivarski:
Yes, I think this is correct—thank you!

PR #518, opened 2020-08-28
Implemented TFileUpdate.bulk_write

bfis:
This PR implements a `bulk_write` for `TFileUpdate`, which is useful for writing O(10^5) objects (i.e. histograms).
In such cases it avoids the performance degradation of the `__setitem__` way of writing, by delaying the update/write of the `TDirectory` entry until after all the `TKey`s have been added. In particular, this reduces accesses of `TKey.fKeylen` from O(n^2) to O(n).

Some details regarding this Implementation:
- currently prevents overwriting entries (existing name+cycle combinations) - but this might actually work fine
- always writes a new `TDirectory` (after all the objects), even if the current allocation is big enough (which is unlikely in the intended usage scenario)
- `TTree` writing was/is not tested

If/When the writing implementation supports delayed `TDirectory` updates this should not be needed anymore.

### DIFF
--- a/tests/test_write.py
+++ b/tests/test_write.py
@@ -2135,7 +2135,7 @@ def test_jagged_i4_manybasket(tmp_path):
     for i, event in enumerate(tree):
         assert(numpy.all([x for x in event.branch] == tester[i]))
 
-def test_bulk_write(tmp_path):
+def test_update(tmp_path):
     filename = join(str(tmp_path), "example.root")
     testfile = join(str(tmp_path), "test.root")
     n = 3
@@ -2148,7 +2148,7 @@ def test_bulk_write(tmp_path):
     t = uproot.open(testfile)
     hist = t["hvar"]
     with uproot.recreate(filename, compression=None) as f:
-        f.bulk_write(("test%d" % i, hist) for i in range(n))
+        f.update(("test%d" % i, hist) for i in range(n))
 
     f = ROOT.TFile.Open(filename)
     for i in range(n):
--- a/tests/test_write.py
+++ b/tests/test_write.py
@@ -2134,3 +2134,22 @@ def test_jagged_i4_manybasket(tmp_path):
     tree = f.Get("t")
     for i, event in enumerate(tree):
         assert(numpy.all([x for x in event.branch] == tester[i]))
+
+def test_bulk_write(tmp_path):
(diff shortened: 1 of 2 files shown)


jpivarski:
This isn't updating the file—it's replacing it, right? And it's not making a subdirectory—it's filling the root directory, right? When I last left off, Uproot couldn't write nested directories.

I might be tempted to call this `update` in analogy with the MutableMapping method that sets a lot of items all at once. But `update` adds new key-value pairs without removing old ones that aren't in the Mapping that is being overlaid on it. Does this do that? If it has or can be made to have the same semantics as `update`, naming it that would make it a more discoverable feature.

I'm in favor of this change and will be merging it. It's also useful to learn that the mechanics involved in assignment are such a bottleneck that they need to be circumvented—maybe something can be done about that when I port then too Uproot 4.

Let me know when you're absolutely done so that I don't merge it before you're ready!

bfis:
> This isn't updating the file—it's replacing it, right?

No, it is updating the file.

> And it's not making a subdirectory—it's filling the root directory, right?

Yes, it is filling the root directory.

> When I last left off, Uproot couldn't write nested directories.

Yes, this is this is still the case.

> I might be tempted to call this `update` in analogy with the MutableMapping method that sets a lot of items all at once. But `update` adds new key-value pairs without removing old ones that aren't in the Mapping that is being overlaid on it. Does this do that? If it has or can be made to have the same semantics as `update`, naming it that would make it a more discoverable feature.

That is a very good idea, I've now changed it accordingly. I've also removed the limitation regarding overwriting existing keys (`name;cycle`) to make it equivalent to `dict.update`.

> Let me know when you're absolutely done so that I don't merge it before you're ready!

I have no further changes planned.

jpivarski:
Okay, then I'll take it as is!

By "replacing, not updating," I mean that we can't add objects (e.g. histograms) to a file created by ROOT. Here, I think you mean that it updates a writable file that Uproot has opened, but Uproot replaces the file when it's opened. Accepting a `TFree` in any state ROOT puts it in is not a problem that we solved, and it would require extraordinary proof: we'd have to create a file in ROOT, add histograms to it, and then add more histograms to the same file in ROOT, maybe with yet another round-trip after that to show that the updates to `TFree` that ROOT and Uproot make are truly compatible.

PR #534, opened 2021-02-12
Update TTree.py

NiclasEich:
Changed the for-loop for the JaggedArray datatype to a version flattening the array. The for loops runtime scaled quadratically and was not possible to use for arrays with more than 1e5 entries.


jpivarski:
That's absolutely a good fix—thank you! I just tested it, and all looks good. I'll release this as a new patch release after fixing some NumPy 1.20 warnings.

